PR #614, opened 2017-02-14
update to typings 1.0

marsmith:
I couldn't find a way to install typings without disabling ssl with `rejectUnauthorized:false` in a .typingsrc file

links:
https://github.com/typings/typings/issues/106
https://github.com/typings/typings/issues/416
https://github.com/nodejs/node/issues/3742


jknewson (approved):
Looks good

PR #633, opened 2017-02-16
v4.0.9

marsmith:


jknewson (approved):
Looks good. I think this is the best version so far.

PR #640, opened 2017-02-17
Cloud fixes

marsmith:
- fix streamgage query
- use square meters value from percent overlay to comput areas in flow report headings
- add equivalent years of record to report
- show parameter code instead of name in parameter table
- fix BC table widths so they are visible


jknewson (approved):
looks good

PR #664, opened 2017-03-15
release 4.1.2

marsmith:
-upgrade dependencies
-fix region rounding
-implement USGS Search API v2
-change search result to circlemarker
-fix 'start over' button to make sure previous results are cleared
-percentOverlay code to split array of values implemented
-set CRB, DRB, RRB to BC only


jknewson (commented):
not sure why but dist/appconfig wasn't updated updated. It still shows 4.1.1, Probably needs to be built again? Also I'm wondering if we should incorporate some of the easy issues before we pull into master, such as #662 etc.

jknewson (approved):

PR #1036, opened 2020-06-03
Mj sdev

marsmith:


src/index.html
```diff
@@ -56,7 +56,6 @@
     <script src="../node_modules/d3/d3.js"></script>
     <script src="../node_modules/nvd3/build/nv.d3.js"></script>
     <script src="../node_modules/angular-nvd3/dist/angular-nvd3.js"></script>
-    <script src="../node_modules/save-svg-as-png/lib/saveSvgAsPng.js"></script>
```
kjacobsen16: This is needed for the report when the QPPQ/FDCTM method is enabled

kjacobsen16 (changes requested):
The streamgages are still displaying on top of the exclusion polygons, since they're both from the national layer.

Edit:  Hopefully we can get it in the app, but let me know if this needs to be fixed in the service,

src/index.html
marsmith: weird , i didnt take that out

maybe my npm lib is old

marsmith:
> The streamgages are still displaying on top of the exclusion polygons, since they're both from the national layer.
> 
> Edit: Hopefully we can get it in the app, but let me know if this needs to be fixed in the service,

exclusion polygons are not in the national layer, they are in the state layers making this unusually complicated, along with the angular leaflet-directive were using not really supported layer ordering. 
  I thought i had this working tho I'll take a look

kjacobsen16:
Oh right, don't know how I got that mixed up (I've been working with that layer a ton this week) 😅 sounds good

USGSPMM3:
I don't want to pile on, but if we end up redoing exclusion polygons, Ryan Thompson put together a document on some suggested changes and revisions we should consider.

Peter McCarthy
National StreamStats Coordinator
WY-MT Water Science Center
Helena, MT
406-457-5934 (w)
406-594-4067 (c)
StreamStats Home page: https://streamstats.usgs.gov<https://streamstats.usgs.gov/>
Follow us on Twitter @USGSStreamStats<https://twitter.com/USGSStreamStats>


________________________________

kjacobsen16:
Thanks Pete - I'll keep that in mind, but hopefully we won't need to adjust the state service that has the exclusion polygons yet (I got mixed up when I said they were in the same layer as the streamgages).

Can you send that document over to me though? We should maybe add it as a future enhancement issue.

kjacobsen16 (approved):
👍 works great, just need to adjust the opacity of the hucs layer (writing this to remind myself)

marsmith:
Or just use a lighter gray? either way

Issue #1040, opened 2020-06-30
Create gage page modal


HansVraga:
Leverage new gage stats services (examples: 
https://test.streamstats.usgs.gov/gagestatsservices/stations/552
    https://test.streamstats.usgs.gov/gagestatsservices/stations/05301200
)

HansVraga:
Example of the existing gage pages https://streamstatsags.cr.usgs.gov/gagepages/html/03572210.htm

kjacobsen16:
@marsmith Pete asked if we can allow a user to filter the statistics shown in the gage pages by statistic group. Can you set that up on the client side? Also, are we displaying the statistic group for the statistics (I know characteristics don't have groups)?

kjacobsen16:
Another ask - add options for "Include all reported statistics" or "include all preferred statistics" (hides stats with IsPreferred = false)

We'll be making these changes in the gagestats management client as well, so whoever gets to it first could share code to speed things up

Issue #1030, opened 2020-04-29
Basin editing issue 


marsmith:
We have an issue where the delineated basin polygon geometry has some issue where adding or removing edit polygons throws a topology error in the console: `found non-noded intersection between LINESTRING` .    Example delineation in ohio shows the error: 41.55993, -83.86761.

kjacobsen16:
@marty reopening this, I'm seeing this issue in another area found by the same user with the first issue.

There's a disconnection of the stream grid downstream from 41.4783857, -83.8293846 at a ditch, so basins upstream of that disconnection are cut off. I can add new items if I delineate just above the ditch at that point, but if I go further upstream (41.48998, -83.80676) and try to add, I get the topology error (see image below).
![image](https://user-images.githubusercontent.com/35497731/83194904-01787e80-a0ff-11ea-91a4-0526c6d5f993.png)

kjacobsen16:
Also worth noting here - I tried to edit several basins in ND and MN and got the same error. I found a point in NY and it worked fine. I wonder if this is a locational issue.

marsmith:
Consider rolling back latest changes on basin editing as it seems like my changes created more issues.  Also, there is still an issue with basin edit ordering on the back end, after the problematic front end edits are submitted.  Possibly respond to the user that 'this is more complicated than we thought.'

Issue #1029, opened 2020-04-29
Issue with citationUrl getting returned from estimate flows 


marsmith:
there was a temp fix in the code that was breaking the citation url.  Fixed that and introduct another error because 'https://' is not included in flows response

Issue #1024, opened 2020-04-01
Test application report warning message


USGSPMM3:
Add a warning or watermark or other notification in the reports coming from the test application that this is a development version, not for public consumption, etc. Currently, a report could be created from the test application and I can't tell that it is using test data, etc.

marsmith:
Added message at bottom of report 
![image](https://user-images.githubusercontent.com/6529123/80618182-fd2e4800-8a10-11ea-8d71-0f5f31a85e80.png)

PR #1046, opened 2020-08-17
Merge dev into staging

### LINKED ISSUE #1021: QPPQ Feedback - Suggestions/Enhancements
Might want to break out into smaller issues or filter things we don't want

- [x] When selecting reference gage from map, show which gages can be used - KJ asked for more info
- [x] ~Will there be the option to enter a reference gage?~ added to new issue
- [x] ~Allow use of MM/DD/YYYY format in choosing dates~ added to new issue
- [x] Add ability to modify date range and run again or grey out if not an option. Currently if a report is generated and then a different date range chosen to rerun scenario, the first date range will be used. Only way to prevent this is to refresh the page, open in a different tab or use the 'Start Over' button 
- [x] Ability to get a summary report of basin characteristics of (and possibly distances to) the potential reference gages (this had a dependency on Gagestatsservices)
  - DA, lat, long are provided after ref gage is selected in the FDC output section of the report, but same info not provided (in that section of report) for ungaged location. Would this be good place for it? 
  - Only ungaged FDC provided, show ref gage FDC too?  Both gaged and ungaged hydrographs provided 
- [x] ~Doing this over a year takes a fair bit longer to build the report - from Jeremy "this has to do with optimizing the retrieval to NWIS flow data. If you select over 2+ yrs of data, it'll slow down the response. I didn't have much time to look into it, I'm guessing it might be client side loading of the timeseries (but it could also be the query/processing of the data server side as well)"~ added to new issue
- [x] ~There is no way to simply exit out of the FDCTM window without either accepting the current reference gage "x Ok" or "Select from Map" which will take you back to the map, where you can then access options in the side panel if desired.~ added to new issue
    - If you have no reference gage selected, "x Ok" won't allow you to exit. Maybe a simple x in the upper corner?

### LINKED ISSUE #1028: Don't query citations in report for scenarios who return no regression regions
This seems like a rare occurrence, but if the scenario returns an empty list for the regression regions of a scenario, the query for citations gets sent through with no regression regions in the filter, returning the entire list of citations.

### LINKED ISSUE #1043: Error handling w/ Prosper service
This might be partially due to #1010, but when the prosper service fails, this toast with the "Loading" message stays there. We should remove that and show an error when it fails.
image

### LINKED ISSUE #1055: QPPQ - date selection resets after gage is selected on the map
I think the date gets reset each time the modal opens, so after selecting a gage it goes back to the default date range

### LINKED ISSUE #1066: QPPQ - Add nearest gage functionality
The nearest gage functionality will be used for QPPQ, but we'll also want to make it available for the rest of the app.

Functionality we want:
- [x] Add button to show nearest gages after delineation - the idea was to query while the delineation is being done, but that wouldn't allow them to specify a search radius or distance on network
- [x] Get the statistics and characteristics for the nearby gages
    - I can't remember from our discussion if we just wanted to add a button to view the gage page of each gage or not. If not, we may need to add an "includeStats" option to the nearest gage calls, or include them on default

- [x] Allow the user to select from the nearest gages for the QPPQ reference gage.

Note: I'm wondering if we can adapt our network navigation tools a bit for this

### LINKED ISSUE #1067: Export to KML/KMZ
Feature request from cooperator: Export to KML/KMZ
```
We often use the feature to export the delineated watersheds for use in GIS – shape files & Geo JONS (?)

Could we add an option to get output in KML/KMZ? This would be wildly popular. In particular, we could then load directly into our DOT mapviewer.

Thanks

Charlie

Charles Hebson, P.E.

MaineDOT / Environmental Office

State House Station 16

Augusta ME 04333-0016

207.557.1052
```

### LINKED ISSUE #1071: SC Roads layer
The SC roads layer was changed at some point and is no longer working. Kitty wants the layer removed.

I patched this into the prod application to prevent users from submitting tickets, but this needs to be adjusted in the repo as well.

### LINKED ISSUE #1075: QPPQ changes from 3/2 call
- [x] More clearly label the FDC plots and tables to make it clear that the charts are for the _ungaged_ location (since the reference gage information is right above the charts)
    - See Mackenzie's notes in Feedback document
- [x] Remove option for Published FDC (want to _only_ be able to use Published FDC)
- [x] Don't allow selection of gage if no published flow duration curve?
- [x] Change estimated flows from QPPQ below 0.005 to 0 in the report table
- [x] ~(awaiting more from Mackenzie/Amy) limit gage selection in IL/IN to gages in approved list~ moved to new issue

Other things KJ and Pete noted
- [x] Show click point drainage area
- [x] ~Do we need to notify if gage is upstream or downstream~ moved to new issue
- [x] Clear selected reference gage when re-querying
- [x] map selection - improvements needed for reference gage requirements/selection
- [x] ~Report print needs Mitch help~ moved to new issue

- Double check 3/2 agenda for anything missed

kjacobsen16:
- Error fixes: leaflet elevation control, misc. dependencies
- Adds npm script (`npm start`) for running project from VSCode w/ auto compiling
- #1023, #1024, #989, #1029, #1019, #1030, #1018, #972 
- Changes to basin editing workflow
- Point to nationalLayers_test map service (changes to huc boundary transparency)
- Updates for IL/IN QPPQ (#1020, #1022)
- Add in new gage page (#1040), #1062
- Fixes #1043

QPPQ Updates
- Fixes #1021 - misc. suggestions/enhancements for QPPQ
    - displays table with available gages for user selection
    - adds additional helpful data, such as drainage area, NWIS start/end date, ability to open gage page, etc.
- Fixes #1066 - nearest gage functionality for QPPQ
    - Queries nearest gages and displays in table for user selection
- Fixes #1075 - QPPQ changes from March call
    - More clearly label the FDC plots to make it clear they are for the ungaged location
    - Don't allow selection of gage if no published flow duration curve
    - Changed estimated flows below 0.005 to 0 in report table
    - Show drainage area of clicked point
- Fixes #1055 - keep selected date selection when QPPQ modal is re-opened

Additional Issues
- Fixes #1028 - don't query citations if no regression regions are returned for a scenario
- Fixes #1067 - add ability to export to kml
- Fixes #1071 - removes SC roads layer from appconfig
- Query streamgages from GageStats DB instead of map service


kjacobsen16:
I noticed an issue in dev where the report isn't displaying the basin chars in some cases - I noticed this in MN when I just selected all basin chars and no scenarios - possibly caused by changes for QPPQ 

- Edit: fixed in KJ-dev branch

harper-wavra (approved):

PR #1070, opened 2021-01-19
Tag#947

tgross-usgs:
Appends temporary water-use registrations (permits) for both Surface Water and Groundwater as well as Indexes with and without temporary registrations to the Wateruse Annual Table. 

Sample click-point at 39.86018, -80.80170 matches values from tab 2 in Jeremy's Excel spreadsheet (attached).
[Web_mockup.xlsx](https://github.com/USGS-WiM/StreamStats/files/5838956/Web_mockup.xlsx)

### DIFF
--- a/src/Controllers/WateruseController.ts
+++ b/src/Controllers/WateruseController.ts
@@ -89,8 +89,8 @@ module StreamStats.Controllers {
         public computeDomesticWU: boolean;
 
         public CanContinue: boolean;
-        public AnnualTotalWithdrawals = 0;
-        public AnnualTotalReturns = 0;
+        private AnnualTotalWithdrawals = 0;
+        private AnnualTotalReturns = 0;
         public MonthlyReportOptions: any;
         public MonthlyReturnReportOptions: any;
         public AnnualReportOptions: any;
@@ -98,7 +98,20 @@ module StreamStats.Controllers {
         public SelectedTab: WaterUseTabType;
         public SelectedWaterUseType: WaterUseType; 
         public ReportData: IWaterUseReportable;
-        public Q10 = 39.5 * 0.646316889697;
+        public get Q10(): any {
+            if(this.NSSService.selectedStatisticsGroupList) {
+                if(this.NSSService.selectedStatisticsGroupList[0]) {
+                    if(this.NSSService.selectedStatisticsGroupList[0].regressionRegions[0]) {
+                        if(this.NSSService.selectedStatisticsGroupList[0].regressionRegions[0].results[0] && this.AnnualTotalReturns && this.AnnualTotalWithdrawals && this.permits_gw.aveWithdrawal && this.permits_sw.aveWithdrawal) {
+                            return +this.NSSService.selectedStatisticsGroupList[0].regressionRegions[0].results[0].value * 0.646316889697;
+                        }
+                    }
+                } 
+            } else { return "---" }            
+        }
+
+        private permits_sw = { name: "Temporary water-use registrations (surface water)", aveReturn: "", aveWithdrawal: "---"};
+        private permits_gw = { name: "Temporary water-use registrations (ground water)", aveReturn: "", aveWithdrawal: "---"};
       //Constructor
         //-+-+-+-+-+-+-+-+-+-+-+-
         static $inject = ['$scope', '$http', 'StreamStats.Services.StudyAreaService', '$modalInstance', '$timeout', 'WiM.Event.EventManager', 'StreamStats.Services.StudyAreaService'];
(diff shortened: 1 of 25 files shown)


src/Services/nssService.ts
```diff
@@ -232,6 +235,10 @@ module StreamStats.Services {
                     //check to make sure there is a valid response
                     if (response.data[0].regressionRegions[0].parameters && response.data[0].regressionRegions[0].parameters.length > 0) {
 
+                        if(this.selectedStatisticsGroupList.length == 0) {
+                            this.selectedStatisticsGroupList.push({'name': "", 'id': "7", 'code': "", 'regressionRegions': [], 'citations': null, 'disclaimers': ""})
+                        }
+
```
kjacobsen16, line 248: I have a couple thoughts:
1. Can you add a comment explaining why this was added for future developers?
2. Does this only get called when (1) a statistic group is selected from the available list in the sidebar or (2) OH WU calls it? I just want to make sure that there won't be any other cases where this gets called and there are no other statistic groups selected where it might result in errors.

src/appConfig.js
```diff
@@ -21,7 +21,7 @@ if (window.location.host === 'streamstats.usgs.gov') {
         configuration.baseurls.StreamStatsMapServices = 'https://gis.streamstats.usgs.gov',
         configuration.baseurls.NSS = 'https://streamstats.usgs.gov/nssservicesv2',
         configuration.baseurls.nssservicesv2 = 'https://streamstats.usgs.gov/nssservicesv2',
-        configuration.baseurls.WaterUseServices = 'https://streamstats.usgs.gov/wateruseservices',
+        configuration.baseurls.WaterUseServices = 'https://test.streamstats.usgs.gov/wateruseservices',
```
kjacobsen16, line 24: This is the section that gives the base urls for prod SS. We should probably push these to the prod wateruseservices when we're ready. Otherwise, it'll get migrated with the next migration anyway.

src/Controllers/WateruseController.ts
```diff
@@ -143,6 +163,21 @@ module StreamStats.Controllers {
...
+            var Q10 = 0;
+            if(this.NSSService.selectedStatisticsGroupList) {
+                if(this.NSSService.selectedStatisticsGroupList[0]) {
+                    if(this.NSSService.selectedStatisticsGroupList[0].regressionRegions[0]) {
+                        if(this.NSSService.selectedStatisticsGroupList[0].regressionRegions[0].results[0] && this.AnnualTotalReturns != null && this.AnnualTotalWithdrawals != null && this.permits_gw.aveWithdrawal != null && this.permits_sw.aveWithdrawal != null) {
+                            Q10 = +this.NSSService.selectedStatisticsGroupList[0].regressionRegions[0].results[0].value * 0.646316889697;
+                            this._indexWOReg = (+this.AnnualTotalWithdrawals - +this.AnnualTotalReturns) / Q10;
+                            this._indexWReg = (+this.permits_gw.aveWithdrawal + +this.permits_sw.aveWithdrawal + (+this.AnnualTotalWithdrawals - +this.AnnualTotalReturns)) / Q10;
+                        }
+                    }
```
kjacobsen16: Hm, if we can I'd like to clean this up. Have you tried putting these on one line, instead of multiple if statements?

src/Services/StudyAreaService.ts
```diff
@@ -982,6 +991,20 @@ module StreamStats.Services {
...
+                    if(e != null && e.parameterLoaded) {                    
+                        this.nssService.estimateFlows(this.studyAreaParameterList,"value", this.selectedStudyArea.RegionID, false, regtype, false);
+                        this._onQ10Loaded.raise(null, WiM.Event.EventArgs.Empty);
+                        //console.log(this.nssService.selectedStatisticsGroupList);
+                    }
+                }));
+                this.loadParameters();
+            }));
+            this.nssService.loadParametersByStatisticsGroup(this.selectedStudyArea.RegionID, "", "GC1449, GC1450", [], regtype);            
+        }
```
kjacobsen16: Is there a reason this (and the Q10Loaded event/subscriber) is in the studyAreaService, rather than the NSSservice?

kjacobsen16 (changes requested):
Overall appears to be working well from what I can tell. I just have some general questions/cleanup suggestions.

We can still work on putting this on a test site for Greg since it's functional. We can set up a call so I can show you how to do that.

src/Services/StudyAreaService.ts
tgross-usgs: The study area service seemed the best place because we're calculating everything for a study area and had actually set up a wateruseQ10 property on the StudyArea model initially. As we neared the end we ended up not populating that property but I'm wondering if it might be a good idea to go back and populate it anyway and then just call it in the wateruse controller.  I guess it feels like we'd be muddying the NSSService to have placed this in there, especially if we're modifying properties in the studyarea.

src/Controllers/WateruseController.ts
tgross-usgs: This was actually a remnant of a previous issue I was having looking into the wrong service - one of those copy/paste and forget to replace the value errors on my part. The properties kept coming up undefined and in the past I've had to do weird checks like this, breaking down each element and checking to see if it exists but now that I think about it that's only happened to me with the new angular. Anyway, just explaining the madness behind this so you don't think I'm crazy! ;-) I was able to shorten it considerably.

src/Services/nssService.ts
tgross-usgs: I added a comment that it's a check for OH WU prior to user scenario selection.  Choosing a scenario in a normal situation sets the selectedStatisticsGroupList.length to 1 right away. With wateruse the user isn't selecting a scenario so this list is unpopulated at the water-use stage of the StreamStats workflow. The important thing is that in a normal non-water-use situation, the if statement always fails, which is what we want.

src/Services/StudyAreaService.ts
tgross-usgs: I populated and now use the wateruseQ10 property in the StudyArea as was originally intended.

tgross-usgs:
I noticed as I did some testing after the first round of comments that the sidebar functionality wasn't working as intended. To remedy this I needed to unsubscribe from several events being raised in the nssService and StudyAreaService. It wasn't straightforward how to unsubscribe from everything the way set up the initial nested subscriptions so I broke the nested subscriptions apart and set each subscription/unsubscription up within new functions. Let me know if you have questions!

kjacobsen16 (approved):
Looks good (at least at the one test point 😅 ).

I think our next step will be putting it on a test site for Greg.

PR #1074, opened 2021-02-25
Tag#947

tgross-usgs:
Improves water-use index (dimensionless) without temporary registrations by computing water use index values with annual average net instead of annual total net value.


kjacobsen16 (approved):
Looks good (from my little knowledge)

PR #1077, opened 2021-03-18
Tag#947

tgross-usgs:
- Fixes errors in water use functionality for non-OH states
- Adds titles to water use pie charts, removes legends from pie charts, and reduces wasted space
- Improves text/spinners on Get Water Use button in water use modal


kjacobsen16 (approved):
Looks good!

PR #1108, opened 2021-04-20
Fix issue when panning streamgage layer

kjacobsen16:
- Fixes issue where, if you zoom in using shift + drag then try to pan, the gages layer didn't update because leaflet wasn't registering a change in the layers

- I had to change where the legend symbology goes in order to get this to work, it's now under Application Layers"


tgross-usgs (approved):
I found where the Streamgages are located now in the legend and they all appear when panning around the map. Looks good to me!

PR #1111, opened 2021-04-20
Merge staging into master

kjacobsen16:
- Fixes freshdesk token issues
- Adds error message for runoff modeling services
- Adds survey link to sidebar footer - see "Tell Us How you use Streamstats" link
- Adds Google Analytics tracking of drainage area and lat/long
- Fixes issues with Streamgages layer switch and legend setup - "Streamgages" layer toggle now works and legend displays symbology


tgross-usgs (approved):
Issues running it locally but that may be my wrong node version... Looks good on test site. :-)

Issue #1125, opened 2021-07-02
Fix removeGeoJson() function in MapController


kjacobsen16:
None of the geojson layers are being removed from the legend when the removeGeoJson() function in the MapController is called.

This is called when a basin is cleared, so can cause some confusion when a basin is cleared if those items aren't on the map. This is especially confusing with the Non-simplified basin, since some basins don't need simplification and don't have a non-simplified polygon to display.
![image](https://user-images.githubusercontent.com/35497731/124295188-2c14e200-db1e-11eb-8152-53027dccd6d8.png)

To recreate:
1. Delineate at 18.44253, -66.36785 in Puerto Rico. You'll receive a warning that the app is displaying a simplified basin. The Non-Simplified basin should show up in the legend under "Application Layers" unchecked. If you toggle it on, you should see a red polygon covering the basin.
![image](https://user-images.githubusercontent.com/35497731/124295630-b0fffb80-db1e-11eb-98ad-5a40dfa711f9.png)

2. Hit the "Clear Basin" option in the sidebar
![image](https://user-images.githubusercontent.com/35497731/124295657-b8270980-db1e-11eb-9ea4-6c75730dc431.png)

3. Check the Legend - you should still see the non-simplified basin, basin and click point in there even though they're not on the map
![image](https://user-images.githubusercontent.com/35497731/124295700-c412cb80-db1e-11eb-81f8-0ab7aa26b9a5.png)

4. Delineate at 18.35679, -66.34300. The non-simplified basin layer will still be checked in the legend, but if you toggle it off/on, nothing shows up because the basin is small enough that it isn't simplified and therefore does not have a non-simplified basin option.
![image](https://user-images.githubusercontent.com/35497731/124295817-e4428a80-db1e-11eb-80bd-5ce10926c98a.png)
5. Clear basin again and delineate at 18.34899, -66.47179. This time, there is a non-simplified basin but the toggle is on, so it looks like there isn't one. If you toggle the non-simplified basin off/on again in the legend, you'll see the non-simplified basin.
![image](https://user-images.githubusercontent.com/35497731/124296595-c1fd3c80-db1f-11eb-8b34-ede6ed4fe92c.png)

harper-wavra:
Should be able to close after https://github.com/USGS-WiM/wim_angular/pull/113 is merged.

amedenblik:
Fixed with https://github.com/USGS-WiM/wim_angular/pull/113

Issue #1114, opened 2021-04-23
Allow display of non-simplified basin


kjacobsen16:
Since the simplification is done client-side, we can add the non-simplified as an optional layer in the legend (off on default)

kjacobsen16:
This is live on the dev site (as a test to see if it would work). Pete M is happy with it, but there is a remaining issue:

- [ ] Non-simplified basin is not clearing when basin is cleared - if I delineate a smaller basin that is NOT simplified, the non-simplified outline for the first basin that was cleared shows up when the non-simplified basin is selected in the legend.

Once that is fixed, we can push this up the pipeline

PR #1126, opened 2021-07-02
Fixes #1102, #1123, #1114, #1124, enable WV for migration

### LINKED ISSUE #1102: Mystic River Customizations
Customizations requested by Pete S., in order of prioritization:
- [x] When a user selects MO_STL or MRB, switch the basemap to imagery
    a. This is to fix the issue of the zoom limits, rather than allowing users to zoom further which would display an empty map beyond the basemap extent
    b. The other option is to add a note in the sidebar telling the user to switch basemaps to zoom in further.
- [ ] Highlight pipe and stream features when the user hovers over them
    a. These are the stormdrain layers, not the stream grid
- [x] Turn the stream grid layer for MRB off by default (but still have it available for the user to turn on)

Pete M. has approved 8 hours to work on these issues.

kjacobsen16:
- #1124 -  fixes issues where the min/max limits are converted from metric to imperial (only an issue in PR flow duration stats) - you can see an example of what the error looked like [here](https://github.com/USGS-WiM/NSSServices/issues/222)
- #1123 
- #1114 - adds a "non-simplified basin" option to the legend, default toggle off (PR is the best place to test this)
    - there are some edge cases that causes issues with this (#1125), but it works great for sessions where a single basin is delineated, which should be most sessions
- #1102 
    - Task 1: add "stormDrain" as an application in the appConfig and switch the basemap to imagery when a stormDrain are is selected (this only will affect MO St Louis for now).
    - Task 3: Turn the stream grid layer for MRB off by default
    - Note that MRB is disabled since it's still in process of getting set up - you can skip MRB-specific tests
- Enable West Virginia in preparation for the July 8th migration


kjacobsen16:
@harper-wavra I pushed this to the dev site: https://dev.streamstats.usgs.gov/ss/ so you don't have to configure your npm versions and all that to test this out.

There are limited states available for testing on the dev site:
image
But PR should get you most of the things that need testing

harper-wavra (approved):
Everything worked for me.

PR #1138, opened 2021-08-03
Fixes #1137

### LINKED ISSUE #1137: Enable US_National/Bieger Equations
In the appconfig.js and change anywhere it says "regions={0}" to "regions={0},NA". This should be in 3 places (statisticsGroupLookup, statisticsGroupParameterLookup, estimateFlows)

If you want to test this, I usually:
log into the NSS-dev site and select US_National in the dropdown
Go to "Edit Scenarios" and click edit next to one of the regression regions and look at the polygon for that region
Find a spot that is within that polygon and delineate it in your local SS application. You should get equations for that regression region output in the report at the end. You can also watch the Network tab to see if the "NA" is included in the calls it's making.

harper-wavra:
In the appconfig.js changed anywhere it says "regions={0}" to "regions={0},NA".
Ran npm build


amedenblik (approved):
Confirmed that US_National/Bieger equations are showing up in the report for the corresponding regions.

PR #1140, opened 2021-08-04
Fixes #1139

### LINKED ISSUE #1139: Mystic River: Turn on StreamGrids by default
Reverse a change that was made in issue #1102.

We have added an exclusion polygon for the greater Mystic River basin, which needs to be visible as default. Because it would be more difficult to turn on the exclusion polygon as default while also turning off the stream grid as default, we are going to turn both layers on as default.

harper-wavra:
Turned on StreamGridsand excluded poly by default.

Can check on https://dev.streamstats.usgs.gov/ss/ using 42.52196, -71.13761.


amedenblik (approved):
Confirmed that Mystic River Basin (MRB) exclusion polygon and stream grid are showing up in the "MRB Map Layers" and turned on by default.

amedenblik:
@kjacobsen16 Just checking-- it is fine to merge this PR? When I run on the dev server, Delineate is not working for MRB but does work for other places (ex. DE). Is this expected?

kjacobsen16:
Hm, from what I understand MRB should be working. 

Try at 42.52196, -71.13761. If that doesn't work, make sure the MRB data is at E:\data and read the workspace folder for your test delin (D:\clientdata)

amedenblik:
ClientData is showing this: file E:\data\mrb\archydro\Aberjona\fdr doesn't exist 

image

kjacobsen16:
OK, I'd sent that to Pete then! Likely a data configuration issue.

amedenblik:
@kjacobsen16 the file appears to be here: E:\data\mrb\archydro\Aberjona\Layers\fdr 
image

Should we just move the fdr folder up one level?

kjacobsen16:
It does belong a folder up, but that Layers folder shouldn't be there (based on the data structure of other regions).

Data can get wonky if you don't move it around a certain way, so I would still point it out to Pete so he can fix it.

amedenblik:
Pete has reorganized the data and basin delineation is now working.

Issue #105, opened 2015-10-14
add some markdown styling


marsmith:
_submitted from StreamStats_
*_First Name: *_marty
*_Last Name: *_smith
*_Email: *_marsmith@usgs.gov
*_Description: *_testing here

PR #1160, opened 2021-11-23
Fixes #1152

### LINKED ISSUE #1152: QPPQ Remaining Enhancements
- [x] Take [list of gages from the krig services](https://github.com/USGS-WiM/krigservices/tree/staging/KrigServices/Assets/Data) correlation matrix and only allow users to select gages from that list as a reference gage. [Related Krig Service Issue #10 ](https://github.com/USGS-WiM/krigservices/issues/10)
      - ~~Need to check if these are greyed out~~ They are not
      - Whatever is easiest (I am thinking an error)
      - Its not ideal, we would ideally want to create an endpoint to return all of the sites for each state, but for now we can have the count equal to 300 to return all sites for a state (https://test.streamstats.usgs.gov/krigservices/krig?region=IA&x=-93.5786&y=42.1041&crs=4326&count=300)
 - [x] Use Flow Duration Curve Transfer Method or FDCTM rather than QPPQ, and add additional info somewhere saying "also known as QPPQ"
 - [x] See if we can get a distance from the target location to the index gage we might need to see an external library like turf.js or something else to get a distance between two lat/longs
     - In the table before you select a gage
 - [x] Check response coming from NLDI service failures to make sure it notifies properly when it's an NLDI service failure
    - if you need to edit the error text, you'll need to do so in the gagestats services, which make calls to NLDI to get the nearest gages on network
 - [x] Change anywhere it says "reference gage" to "index gage"
 - [ ] See how long it would take to display whether the gage is upstream or downstream
    - This would also be in the gagestatsservices - I started looking into this at one point - I'll see if I have any changes locally for this when I'm fully back in
    - See [GageStats Issue #105](https://github.com/USGS-WiM/GageStatsServices/issues/105)
  - [ ] ~~This wasn't talked about in this meeting but the last one - they have multivariable regions that have flow duration equations for IL/IN. We removed the polygons because we didn't have the review v. approved status yet. We should get those polygons added back into the DB and make sure the equations are marked as review for now~~
    - Amy/Mackenzie need to get these to Ryan to add to the database
 - [x] Alter “Query Nearest Gages” blurb to say “...within a distance on the network (checks all contributing upstream reaches and all reaches downstream that target location contributes to).” 

Notes:

- They will likely have some report changes coming and they're planning to come back with new text to explain what distance on network means.
- Additional information Issue #1078

harper-wavra:
- Uses a returned list of gages from the krig services to only allow users to select gages from that list as a reference gage. Its not ideal, but I used an existing endpoint and made the count equal to 300 to return all sites for a state.
- Changed "QPPQ" text to "Flow Duration Curve Transfer Method" or "FDCTM"
- Added additional info in the configure FDCTM modal saying "also known as QPPQ"
- Changed "reference gage" to "index gage"
- Altered “Query Nearest Gages” blurb to say “...within a distance on the network (checks all contributing upstream reaches and all reaches downstream that target location contributes to).”
- Added error if user selects non continuous gage from map

In order to test you will need to add "FDCTM" to the IA, IN, and IL applications in the appconfig

This issue also asked to see how long it would take to display whether the gage is upstream or downstream, I think since that is a gagestatsservice change we can create a new issue for that in the correct repo.


src/Views/extensionview.html
```diff
@@ -16,7 +16,7 @@ <h5>Select/Enter a date range.<small style="color:red">&nbsp;&nbsp;(Required)</s
             </div>
         </form>
         <div ng-if="vm.selectedReferenceGage" class="form-group">
-            <h5>Select/Enter a reference gage.<small style="color:red">&nbsp;&nbsp;(Required)</small></h5>
+            <h5>Select/Enter a index gage.<small style="color:red">&nbsp;&nbsp;(Required)</small></h5>
```
amedenblik: change "a index gage" to "an index gage"

amedenblik (changes requested):
Confirmed the following. Please see additional review comment.

✔ Uses a returned list of gages from the krig services to only allow users to select gages from that list as a reference gage. Its not ideal, but I used an existing endpoint and made the count equal to 300 to return all sites for a state.

✔ Changed "QPPQ" text to "Flow Duration Curve Transfer Method" or "FDCTM"

✔ Added additional info in the configure FDCTM modal saying "also known as QPPQ"

✔ Changed "reference gage" to "index gage" 

✔ Altered “Query Nearest Gages” blurb to say “...within a distance on the network (checks all contributing upstream reaches and all reaches downstream that target location contributes to).”

✔ Added error if user selects non continuous gage from map

"This issue also asked to see how long it would take to display whether the gage is upstream or downstream, I think since that is a gagestatsservice change we can create a new issue for that in the correct repo." -- this already exists in [GageStats Issue #105](https://github.com/USGS-WiM/GageStatsServices/issues/105)

amedenblik (approved):

PR #1172, opened 2022-01-11
Fixes #1170 - Add South Carolina dams layer 

### LINKED ISSUE #1170: Add South Carolina dams layer
Add South Carolina dams layer: https://www.sciencebase.gov/catalog/item/6154584dd34e0df5fb9d3a6c

amedenblik:
Closes #1170 

- Added SC Regulation layer to StreamStatsMapServices
- Added Regulation Points client map layer for SC 
- Enabled Regulation application for SC
- Edited MapController so that when there are multiple map layers for a region in the appConfig, the correct popup appears for the layer that was clicked on the map


harper-wavra (commented):
Looks nice! The regulation seemed to work for me but I am sure Kitty will be able to test that better.

A couple things about the popups
- Occasionally a dam popup would be empty, but sometimes if I clicked that same dam a second time it would be populated with data. 
[Empty popup](https://user-images.githubusercontent.com/59063181/148975725-6649389f-3622-4552-be60-9727c9528146.png)
[Empty popup]](https://user-images.githubusercontent.com/59063181/148976143-d95d3b8d-84e9-4f39-8c88-4d94816b2b71.png)
- It also seemed like after regulation was done even less of the dam pops were displaying data in them. 



Maybe we should rebuild before you merge, so its ready for Friday.

amedenblik:
I clicked 50 SC dams and did not encounter any issues with blank pop-ups. Then I delineated a basin, checked for upstream regulation, and clicked another 50 dams and did not encounter any issues with blank pop-ups. I tried this in both Chrome and Edge. I cannot recreate the issue, but it seems to happen infrequently enough to ignore?

harper-wavra:
Maybe a screen share could be helpful, I might be doing something wrong. I just clicked on 10 and 8 of them were empty in SC.

harper-wavra (approved):
Tested 20 dams before regulation and 20 after and all looks good.

PR #1185, opened 2022-02-15
Fixes #1178 - incorrect coordinated discharge stats

### LINKED ISSUE #1178: incorrect coordinated discharge stats
It looks like we have something squirrelly going on with Indiana's coordinated reach functionality. A delineation on the St. Joseph River at 41.72027, -86.26709 is producing a correct watershed, but somehow it is selecting the wrong coordinated stream reach (Judy Creek) and therefore is computing peakflows using the wrong equations. In the screen snip below, the watershed is in pale green, the St Joseph River coordinated reach is the darker blue line, and the Judy Creek coordinated reach is the lighter blue highlighted line. It is about 2800 feet between the confluence of these two coordinated reach lines and the watershed delineation point, so it seems like that should be far enough away from the confluence that it wouldn't be a snapping issue. 

image

I looked into how to code selects a coordinated reach a bit this afternoon. It queries this map service: https://gis.streamstats.usgs.gov/arcgis/rest/services/coordinatedreaches/, with a bounding box. The request for the point 41.72027, -86.26709 looks like https://gis.streamstats.usgs.gov/arcgis/rest/services/coordinatedreaches/in/MapServer/0/query?geometry=-86.26664545448304,41.71337381856141,-86.26754376976012,41.72717167816403&geometryType=esriGeometryEnvelope&inSR=4326&spatialRel=esriSpatialRelIntersects&outFields=eqWithStrID.BASIN_NAME,eqWithStrID.DVA_EQ_ID,eqWithStrID.a10,eqWithStrID.b10,eqWithStrID.a25,eqWithStrID.b25,eqWithStrID.a50,eqWithStrID.b50,eqWithStrID.a100,eqWithStrID.b100,eqWithStrID.a500,eqWithStrID.b500&returnGeometry=false&returnIdsOnly=false&returnCountOnly=false&returnZ=false&returnM=false&returnDistinctValues=false&f=pjson. In this case it returned two reaches, Judy Creek and St. Joseph River. It looks like the [code then just takes the first reach](https://github.com/USGS-WiM/StreamStats/blob/5d7de92ac99eb48cd90c757e883361fa2205ab4d/src/Services/StudyAreaService.ts#L741) and that is what it returns. 

It seems like we could try "tightening up" the bounding box used for reach selection. I used the coordinates from the "query?geometry=" portion of the request, just to reconstruct the bounding box used for selection. It is pretty tight from east to west, but strangely stretched out north to south.

amedenblik:
Closes #1178 

The bounding box was incorrect, seemingly due to issues with Leaflet version 0.7.7. Upgrading the Leaflet version caused major changes and downgrading the Leaflet version did not fix the issue.

I scrapped the Leaflet methods and switched to using the TurfJS destination method to calculate the bounding box values 5 meters away from the pour point.


harper-wavra (approved):
Tested the problem sites from the ticket and they all worked! I also checked the bounding box for each as well and they were all a square.  Rebuilt, now I will add to test, and let Ryan know.

PR #1197, opened 2022-02-19
Fixes #1196 - Remove NSSServicesV2

### LINKED ISSUE #1196: Remove NSSServicesV2
Switch https://streamstats.usgs.gov/nssservicesv2 to https://streamstats.usgs.gov/nssservices

https://github.com/USGS-WiM/StreamStats/blob/8d00c48bcd71f106864189e00bba641350953eaa/dist/appConfig.js#L21

amedenblik:
Closes #1196

Removing use of NSSServicesV2 because it is no longer needed.


harper-wavra (approved):
Looks good, and I don't see any other instances of nssservicesv2. 

Due to this [if statement](https://github.com/USGS-WiM/StreamStats/blob/8d00c48bcd71f106864189e00bba641350953eaa/src/appConfig.js#L18-L26) the change will not show up until on production. Dev and test however are both using NSSServices and both look good.

PR #1208, opened 2022-03-08
#1206 - replace watershed boundary source; add HUC 14 and 16

### LINKED ISSUE #1206: Try using the national WBD map server
Currently in the StreamStats client, we display watershed boundaries that are Hydrologic Unit Codes (HUCs), 2-digit through 12-digit: 

image

These watershed boundaries are displayed from our StreamStats GIS Server: https://gis.streamstats.usgs.gov/arcgis/rest/services/StreamStats/nationalLayers/MapServer/1

An example request to display an image of the watershed boundaries is: https://gis.streamstats.usgs.gov/arcgis/rest/services/StreamStats/nationalLayers/MapServer/export?bbox=-9136288.679842262%2C3968686.945594635%2C-9013989.43458598%2C4034957.8491178835&size=1600%2C867&dpi=96&format=png8&transparent=true&bboxSR=3857&imageSR=3857&layers=show%3A1%2C2%2C3%2C4%2C5%2C6%2C7&f=image 

We would like to instead try using this external service to display the watershed boundaries: https://hydro.nationalmap.gov/arcgis/rest/services/wbd/MapServer

This map service will display HUCs that are 2-digit to 16-digit. We would like to use similar symbology to the current map service.

Abigail770:


harper-wavra (commented):
Its working for me! 
I wonder if we should ask Pete to give us labels for the 14 and 16 digit HUC, just so they are consistent? 
They also seems little hard to see, but I think you used the same color that was previously used and it could just be a me problem but maybe if Andrea agrees we can darken it a bit.

I think once Andrea reviews we can push it to dev and let the GIS folks have a look before we push to production.

src/appConfig.js
```diff
@@ -540,15 +540,260 @@ configuration.regions = [
...
+                    "source":{"type":"mapLayer","mapLayerId":1},
+                    "drawingInfo":{
+                        "renderer":{
+                            "type":"simple",
+                            "label":"",
+                            "description":"",
+                            "symbol":{
+                                "color":[0, 0, 0, 0],
+                                "outline":{
+                                    "color":[156, 156, 156, 255],"width":3.0,"type":"esriSLS","style":"esriSLSSolid"
```
amedenblik: I agree with Harper that we should slightly darken the color for all HUCs.

src/appConfig.js
```diff
@@ -540,15 +540,260 @@ configuration.regions = [
...
+                "layerName": "Subwatershed",
+                "layerId": 6,
+                "legend": [{                        
+                    "contentType": "image/png",
+                    "imageData": "iVBORw0KGgoAAAANSUhEUgAAABQAAAAUCAYAAACNiR0NAAAAAXNSR0IB2cksfwAAAAlwSFlzAAAOxAAADsQBlSsOGwAAAElJREFUOI3t1DEKACAMA8AI/jTvylvrVEFwqpmkmTpdM2XCnAkAkuIVIjk26IikIDkOML9UsLxtDRtssMGPQccu2hoeA1vdwVsWQMARe7k7EAEAAAAASUVORK5CYII=",
+                    "label": "12-digit HU"
+                }]
+            },
+            {
+                "layerName": "",
```
amedenblik, line 779: We have contacted Pete for guidance on the layerName for HUC-14s and HUC-16s. Will let you know when we have an answer!

amedenblik (approved):
Pete says the HUC-14 and HUC-16 titles can be left blank. The darker color looks good!

Issue #1187, opened 2022-02-15
Format Culverts Popup


harper-wavra:
When you select a culvert point from the map, a popup should be display.

Display/format of the popup is TBD.

HansVraga:
Until we receive guidance/mock-ups of what to show, I would pull the unique identifier of the culvert from the services and simply display that - showing we have capacity to display info from the map services in the pop-up.

PR #1212, opened 2022-03-10
1186 - add Culverts layer

### LINKED ISSUE #1186: Add Culverts WMS Layer
Add MA Culverts Layer: https://usgs.maps.arcgis.com/home/item.html?id=769d2098dbb34398a6d32ba48e418e44#overview

When a user selects MA the culverts layer should be displayed. As of right now you need to log in to see the map layer. If needed we can have a simple login functionality in order to view layer, could also have a secrets file with login information.

Abigail770:


harper-wavra (approved):
Looks good, I can see the stream crossings in MA.

I can't click on them yet, but I think that is to be expected. This is a good starting point and I can get started on creating a new dev site to host this after Andrea looks at it.

amedenblik (approved):
Looks good to me too! Like Harper said, nothing happens when I click on the points, but I think that is #1187. 

Also adding a note here to remind us that a culverts_secrets.json file is required before the layer is published, and it is available in SharePoint.

PR #1213, opened 2022-03-10
Fixes #1210

### LINKED ISSUE #1210: Handle -9999 equation reponse
- [x] When an equation can not be calculated it return as a -9999 from the services. We need to handle this for the report and print "Undefined" instead of -9999.

related to https://github.com/USGS-WiM/NSSServices/issues/247

harper-wavra:
If equation value comes back from the services as -99999 is will print out as 'undefined' 

Use with PR https://github.com/USGS-WiM/NSSServices/pull/249


amedenblik (changes requested):
The -99999 -> "undefined" change looks good!

The FDCTM section still appears when FDCTM fails to compute. This section should be hidden when FDCTM cannot be computed, and an error message should appear in its place.

image

amedenblik (approved):

PR #1225, opened 2022-03-16
1095 - add timeout to prosper error popup

### LINKED ISSUE #1095: Add timeout to prosper error popup
Right now it stays up until you click it. We should remove it after a certain amount of time.

Abigail770:


harper-wavra (changes requested):
This works well right now. I think we should maybe double the time just so it stays on a bit longer.

PR #1244, opened 2022-03-22
#1228 - update culverts data dict and adjust for new param codes in WMS

### LINKED ISSUE #1228: Update report/data dictionary to match updated WMS
Luke updated the param codes in the WMS, so the json file/report format will need to be changed accordingly.

Abigail770:
Gardner has an internal meeting tomorrow and would like to show the updates on the dev site.


harper-wavra (approved):
Looks good. I rebuilt and will push to dev site.

PR #1246, opened 2022-03-22
Fixes #1245

### LINKED ISSUE #1245: Report wont open
Its throwing the error here: https://github.com/USGS-WiM/StreamStats/blob/dev/src/Services/nssService.ts#L452

harper-wavra:
When trying to open the report sometimes an error would be thrown here, https://github.com/USGS-WiM/StreamStats/blob/dev/src/Services/nssService.ts#L452, saying `statGroup.disclaimers` is undefined so it is unable to read the error. I added a line to check for `statGroup.disclaimers` before it tries to read the `statGroup.disclaimers['Error']`.


amedenblik (approved):
Works well!

PR #1247, opened 2022-03-23
Dev into Staging

harper-wavra:
Fix for report not opening


harper-wavra:
Waiting until coordinated reach fix is added

PR #1203, opened 2022-02-28
Fixes #1202 - Show error message when QPPQ index flow gage values are invalid

amedenblik:
Closes #1202

Client fixes associated with https://github.com/USGS-WiM/NSSServices/issues/245

Use this appConfig file for testing with FDCTM application enabled + local services [appConfig.zip](https://github.com/USGS-WiM/StreamStats/files/8155527/appConfig.zip)

PR #1257, opened 2022-03-29
Fixes 1256

harper-wavra:
Reverted https://github.com/USGS-WiM/StreamStats/pull/1203/files and https://github.com/USGS-WiM/StreamStats/pull/1213/files, which included a bug.
Updated version number


amedenblik (approved):
Tested with all problematic points/statistics that were reported on FreshDesk.

PR #1271, opened 2022-04-01
#1262 - add culvert basin disclaimer; restyle citations in culvert report

### LINKED ISSUE #1262: Add disclaimer message about culvert basins in report
Add disclaimer message in the report that states that the basins that is shown (and can be downloaded) is delineated from Lidar, not by StreamStats. Ask Gardner and others to review the statement.

Note: this will only be used for this stage 1 version of Culverts, which does not delineate watersheds on the fly.

Abigail770:
Disclaimer message add below the map: "The basin presented on this map and available for download was delineated from LiDAR and not by StreamStats."

### DIFF
--- a/src/Controllers/SidebarController.js
+++ b/src/Controllers/SidebarController.js
@@ -426,7 +426,6 @@ var StreamStats;
                             }
                         }
                         ;
-                        console.log(paramList);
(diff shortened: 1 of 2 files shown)


amedenblik (approved):
Looks good!

PR #1280, opened 2022-04-11
1264 - replace param codes with param names column; add param description tooltips

### LINKED ISSUE #1264: Culvert report field updates
- [x] Replace param code with param name in tables
- [x] Remove param description column and and replace with tooltip

Wait for feedback from cooperator before completing.

Abigail770:


amedenblik (changes requested):
Looks good! The Town tool-tip appears to be empty. Is it possible to remove this tooltip? Or if that will be a pain, you could update the description to just be "Town" so it's not empty.

amedenblik (approved):

PR #1298, opened 2022-04-28
Fixes #1275

### LINKED ISSUE #1275: Add Maine Mean August Baseflow WMS layer
Add Mean August Baseflow WMS layer on Dev for Maine

https://streamstats.freshdesk.com/a/tickets/117330

https://usgs.maps.arcgis.com/home/item.html?id=eb8d1922d4984488a0c283933504fee6

harper-wavra:
src/appConfig.js - added layer to appconfig
src/Controllers/MapController.ts - 
- Checks when region layers are added, if layer MeanAugustBaseflow is added then we need to update it 
- Checks for when map zooms, if selected region is in ME and zoom is above 11 we need to update the MeanAugustBaseflow layer
- Updates the layers legend and feature color based on zoom level


amedenblik (changes requested):
- [x] 1. Sometimes the pop-up is showing a title of "undefined" instead of "Mean August Baseflow" and shows all properties instead of the ones defined in the queryProperties

image

- [x] 2. The pop-up shows the variable names instead of the nicer long names in the queryProperties 

image
 - [x] 3. 0 should be No, 1 should be Yes for Drainage Area out-of-bounds, Mean July Precip out-of-bounds, % Aquifer Area out-of-bounds, Regulated stream/river.

amedenblik (approved):
🎉

PR #1300, opened 2022-04-28
Fixes #1195

### LINKED ISSUE #1195: Separate Exclude Poly and Stream grid
It has been requested to separate the exude poly and stream grid layers so they can be turned on and off separately.

If we complete this issue, it could be helpful for issue #1182

harper-wavra:
@Abigail770 did 99.99% of this work. I just updated the wim_angular link in the appconfig to use the the new [master_streamstats](https://github.com/USGS-WiM/wim_angular/tree/master_streamstats) branch. 

I think you will need to do an `npm install` before you review to update your local node_modules folder.


amedenblik (approved):
Nice work, @Abigail770! I know some folks will be really excited about this update.

PR #1307, opened 2022-05-11
Fixes #1306

### LINKED ISSUE #1306: QPPQ Results
When QPPQ results are calculated, if there is more than one region, we want to be able to display both multivariate regions & DA only region results. We could create tabs for each region with results displayed under each tab. We want to Default to multivariate regions results.

harper-wavra:
Display tabs for each region with QPPQ results defaulted to show the multivariate regions results. If there is only one region with results (like in IA) there should be no tabs.

For testing you will need to your local update appconfig:
- 'StreamStatsServices': 'https://**dev**.streamstats.usgs.gov'
- add "FDCTM" to IN


src/Controllers/ReportController.ts
```diff
@@ -154,6 +155,14 @@ module StreamStats.Controllers {
             this.sectionCollapsed = [];
             this.basinCharCollapsed = false;
             this.collapsed = false;
+            this.selectedFDCTMTabName = "";
+            if (this.extensions[0].result.length > 1) {
+                this.extensions[0].result.forEach(r => {
+                    if (r.name.toLowerCase().includes("multivar")) {
```
amedenblik, line 164: Do you think we should add an `if` statement to limit this "multivar" distinction to just IN and IL for now? It won't make a difference for now but may reduce unintended behavior for future QPPQ additions.

amedenblik (commented):
Looks fantastic! 🤩 I added one comment for consideration.

src/Controllers/ReportController.ts
harper-wavra: Added comment for future reference

amedenblik (approved):

Issue #1332, opened 2022-06-08
FDCTM plot scale incorrect after section collapse/expand


amedenblik:
Fix issue with FDCTM plot x-axis getting smaller after the FDCTM section is collapsed and expanded.

amedenblik:
Marking as `wontfix` for now due to difficulty resolving the issue. We can reopen, if users report issues.

Issue #1331, opened 2022-06-08
New Release 4.10.0


harper-wavra:
- [x] Update version numbers in the code
- [x] Update CHANGELOG.md with all PR since last release
- [x] Rebuild
- [x] Pull request this branch -> dev -> staging -> master
- [x] Create release in Github with new version number

PR #1326, opened 2022-05-26
Fixes #1313

harper-wavra:
- Shows longest flow path layer automatically on the main map if returned from the services.
- Added loader and disabled download button until longest flow path is downloaded
- Fixed bug that was causing the longest flow path layers to not show up as options in the report map legend if report was opened before longest flow path was done downloading 

Can get testing points from the Elevation Comparision.xlsx Pete sent us a couple weeks ago.

Issue #1261, opened 2022-03-29
Add FDCTM error messages


amedenblik:
Unrevert and fix https://github.com/USGS-WiM/StreamStats/pull/1203/files and https://github.com/USGS-WiM/StreamStats/pull/1213/files.

These PRs had an issue that caused flow statistic reports to not appear if they had error/info/warning messages returned in the services.

harper-wavra:
They caused the statistics to not be calculated - These are being returned by the services, so something in the new code messed this up. Getting the below error at line https://github.com/USGS-WiM/StreamStats/blob/dev/src/Services/nssService.js#L229. The whole header message is {"info":["Low_Flow_LatGT_41.2_wri02_4068 removed; limitation criteria: LAT_CENT>41.2","Count: 1"]}, so when we split by ; it is trying to JSON.parse just {"info":["Low_Flow_LatGT_41.2_wri02_4068 removed; which is an invalid JSON. Not entirely sure how to fix that though and keep what we were trying to do in https://github.com/USGS-WiM/StreamStats/pull/1203/files.

![image](https://user-images.githubusercontent.com/59063181/160632950-34ab00c9-57c8-464d-a527-609508bc6e05.png)

harper-wavra:
We should also add an error message when the services don't return results. One example of this happening is in IA when basin characteristic RSD is undefined, see point 41.26497, -94.33093
![image](https://user-images.githubusercontent.com/40237491/182483237-80f08e0a-bc9d-449e-bdee-59a3ac83e49e.png)

PR #1333, opened 2022-06-08
Fixes #1292 - QPPQ Additions

### LINKED ISSUE #1292: QPPQ Additions
Feedback on https://dev.streamstats.usgs.gov/ss-test-qppq/

1. - [x] The downloadable CSV includes the data for the tab that the user is on (DA-only vs multivariate), which is great. It looks like when the CSV is opened, there is no indication of which method was chosen within it . Would it be difficult to add a cell somewhere that defines which method the user downloaded?
2. - [x] Make the QPPQ section in the report collapsible
        - [ ] Fix issue with plots getting smaller after being collapsed and opened  --> #1332 
        - [x] Fix issue with tab formatting looking weird: 
image Used to look like this: 
image

3. - [x] Add citation to the QPPQ section in the report: https://pubs.er.usgs.gov/publication/sir20155181
4. - [x] The flow-duration statistics are returned as part of the report, in a collapsible section titled the "Flow Duration Curve Transfer Method". Would it be possible to update the title of this section to match the terminology returned form the development site for the FDCs, “Flow-Duration Statistics”? That would help differentiate it from the following section as well (also titled FDCTM). Solution: When FDCTM scenario is selected, automatically select the FDS scenario and do not allow it to be unselected. The report will then contain the FDS section and the FDCTM graph section. The report will not contain an FDCTM statistics section.
     - [x] Fix resulting bug: 
image

5. - [x] The estimated FDC table that is included in the FDCTM output (starts at 0.01 and ends at 99.9) is in the opposite order of estimated FDC in the flow-duration stats section (starts at 99.9 and ends at 0.1). Would it be possible to put those in the same order?
6. - [x] ~Omit the second "(FDCTM)" in the title of the section~ Cannot reproduce this issue
image

amedenblik:
Closes #1292

Use appConfig.js from the https://dev.streamstats.usgs.gov/ss-test-qppq/ site for testing.

1. The downloadable CSV now includes a line ("Regression Region:") to indicate which Regression Region's results are being shown in the FDCTM section.
2. The FDCTM section in the report is now collapsible.
3. Fixed an issue with tab formatting that resulted from making the FDCTM section collapsible.
4. Added citation to the FDCTM section in the Report
5. When the "Flow-Duration Curve Transfer Method" scenario button is selected, the "Flow-Duration Statistics" scenario button is now selected by default. The "Flow-Duration Statistic" scenario button cannot be unselected while the "Flow-Duration Curve Transfer Method" scenario button is selected.
6. The "Flow-Duration Curve Transfer Method" statistics group does not show up in the Report anymore. Instead, the same information appears in the "Flow-Duration Statistics" statistics group.
7. The exceedance probabilities table is now sorted from greatest to least exceedance.

Note: we forgot to update the ChangeLog since the last release. It will be updated with #1331.

Remaining work:
#1332


harper-wavra (changes requested):
I tested out points IA, IN, using QPPQ and other scenarios (couldn't get IL to work the basin chars were not calculating.) Then also some points in other states without QPPQ.

- [x] Points in IA (42.1451735, -92.54651) seem to not be working when I try to Open Report.
[example point and error](https://user-images.githubusercontent.com/59063181/172862061-dcf5fdc2-2d1e-48f6-ae29-f710e4d3e0bf.png)
- [x] Seems like somewhere this [line](https://github.com/USGS-WiM/StreamStats/blob/6c0fa779c20f89c7322bb7246eb34818a47c4876/src/Controllers/ReportController.js#L48) got removed, its causing errors if a longest flow path is present. Not sure where this got removed at just noticed it when reviewing this. I was testing at point 40.488443673323964, -85.67108631134033, selecting Peak Flow Statistics and Flow Duration Statistics, once I added that line back it seemed to work again. It also seems to be downloading LFP even if its not selected now. Something got messed up with that PR, we should probably check to make sure everything from that PR (https://github.com/USGS-WiM/StreamStats/pull/1326) is still in dev/this branch.  Let me know if you don't want to do that in this PR and I can look at it. It was just a bit confusing to test this PR without adding that line back.

harper-wavra:
I just check dev with PR https://github.com/USGS-WiM/StreamStats/pull/1326 and it was all in dev, so it seems like somehow this branch did not get all the changes from the PR.

amedenblik:
Ready for re-review. 

>  it seems like somehow this branch did not get all the changes from the PR

I had to manually resolve merge conflicts right near the line that was inadvertently removed. I must have accidentally deleted it.

harper-wavra (changes requested):
Looks good! A couple things I noticed in the code:
- [x] fix appconfigs
- [x] I think you need to rebuild after your last commit 
- [ ] IL point with no results (maybe just related to https://github.com/USGS-WiM/StreamStats/issues/1261)
     - 39.46730912048314, -85.62623977661133

amedenblik:
Ready for re-review. 

Could not recreate issue with 39.46730912048314, -85.62623977661133.

harper-wavra (approved):

PR #1349, opened 2022-06-29
Fixes #1347 - Handle out of range values in SC Runoff Models

### LINKED ISSUE #1347: Handle out of range values in SC Runoff Models
- [x] 1. For the Bohman Urban method, all values entered by the user must be greater than 0. Total impervious area should be between 0 and 100. If any value does not meet this criteria, when the user clicks the "Continue" button, notify the user and force them to correct the values before moving on.
- [x] 2. Some values and combinations of values may cause a "warning" to appear in the response headers. This warning should be displayed in the report. (~There is a current bug with the service (https://github.com/USGS-WiM/SC-RunoffModelingServices/issues/22)~ resolved).

amedenblik:
Closes #1347 

Must complete this and redeploy first: https://github.com/USGS-WiM/SC-RunoffModelingServices/pull/32

Some testing points:

Only Region 3
33.47655626699917, -81.9091272354126

Only Region 4
33.05054, -80.85092

Both Region 3 & 4
33.87821873743382, -80.68458080291748


amedenblik:
https://github.com/USGS-WiM/SC-RunoffModelingServices/pull/32 has been deployed, so this is ready for review

harper-wavra (changes requested):
Do you know of any example points or values that shouldn't give a warning?

Looks good, just a couple thoughts, feel free to take or leave them.  

I can still continue if some of the values are not in the specified range. I was thinking if the values entered were out of range there would be a little message or warning and the user would not be able to click Continue. I see the error after I click Continue but maybe it makes sense to have some form validation and set a min and max value for those html inputs? That way the user gets the error immediately and doesn't have to wait through all the service calls.

Do we want the warnings to show up in the csv download?

amedenblik:
Ready for re-review. Updates:
1. Validated inputs within the form
2. Added warning to CSV download

The warning messages will show up for every result, because this message will always appear: "These methods are not applicable in basins with large rural sub-basins or areas with extreme contrasts in level of urbanization. "

PR #1356, opened 2022-07-07
Fixes #1354 - Add stream name and HUC8 to South Carolina report

### LINKED ISSUE #1354: Add stream name and HUC8 to South Carolina report
For the South Carolina application:

- [x] When a user clicks a delineation point, find the closest NHD stream line from https://hydro.nationalmap.gov/arcgis/rest/services/nhd/MapServer/6
- [x] When a user click a delineation point, find the HUC8 watershed it is located within https://hydro.nationalmap.gov/arcgis/rest/services/wbd/MapServer/4
- [x] Add these attributes to the report: GNIS_NAME of stream, GNIS_ID of stream, huc8 of watershed, name of watershed
- [x] Add new output to CSV
- [x] Remove toasts besides errors
- [x] Update Changelog
- [x] Test at a variety of locations
     - 33.16685, -80.73001: large stream named on basemap; successfully returned both stream and HUC8 ✅
     - 34.01382, -80.99085: small named stream on basemap; successfully returned both stream and HUC8 ✅
     - 34.31315, -81.80690: not a named stream on basemap; only returned HUC8 
     - 34.31690, -81.82357: tributary to named stream on basemap; only returned HUC8
     - 34.31765, -81.82383: small named stream on basemap; successfully returned both stream and HUC8 ✅
     - 34.31683867381465, -81.82391881942749: small named stream on basemap; only returned HUC8

amedenblik:
Closes #1354 

Doesn't always return NHD streams due to differences in streamgrid and NHD stream lines. We may need to adjust the search tolerance, which is currently 50m. Will get additional feedback from SAWSC and SCDOT after we put this on the Dev site.

Hint: the NHD and WBD data is printed to the console after you click "Continue" after the basin is delineated.
image


harper-wavra (approved):
Seems to be working well for the dev site and to have them review.

Good idea making it is own application! If we decide to go with the national map service, it would be easy to add this to other specific states. Then if we decide we want this for all states it would be easy enough to get rid of the separate application and just make it a part of the report for all states.

Issue #1385, opened 2022-07-22
SC Synthetic UH: Calculate parameters in user interface


amedenblik:
See parent issue https://github.com/USGS-WiM/StreamStats/issues/1344 for more details
Complete #1384 first

At the bottom of the user interface will be a "Calculate Missing Parameters" button. When the user clicks that button, the following values should be calculated and entered into the user interface. See the "USGS SC Flood Hydrograph for Urban Watersheds" tab to get an idea of how this works. 

Some of these values will be typical StreamStats Basin Characteristics, which can use the `calculateParameters` function in `SCStormRunoffController.ts` as it is currently written. Other values will need to be retrieved via the `calculateMissingParametersSCSUH` endpoint in the [SC-RunoffModelingServices](https://github.com/USGS-WiM/SC-RunoffModelingServices/), so the `calculateParameters` function will need to be edited to accommodate those.

You can test out the `calculateMissingParametersSCSUH1` endpoint here: https://sc-runoffmodelingservices.streamstats.usgs.gov/docs#/default/calculatemissingparametersSCSUH_calculatemissingparametersSCSUH___post. The example input is provided for the Travel Time Method. Example input is provided for the travel time method. You can click on the "Schema" tab to view the other inputs that would be relevant for the Lag Time Equation.

Values to be calculated:

* Rainfall Distribution Curve: use the "rainfall_distribution_curve_letter" value from the `calculateMissingParametersSCSUH` endpoint in SC-RunoffModelingServices
* Drainage Area (acres): StreamStats basin characteristic DRNAREA
* Time of Concentration, Tc (minutes): use "value" in the "time_of_concentration" value from the `calculateMissingParametersSCSUH` endpoint in SC-RunoffModelingServices
* Peak Rate Factor, PRF: use the "peak_rate_factor" value from the `calculateMissingParametersSCSUH` endpoint in SC-RunoffModelingServices
* Standard Curve Number, CN: use value "curve_number" from the `calculateMissingParametersSCSUH` endpoint in SC-RunoffModelingServices
* Watershed Retention, S: use value "S" from the `calculateMissingParametersSCSUH` endpoint in SC-RunoffModelingServices
* Initial Abstraction, Ia:  use value "Ia" from the `calculateMissingParametersSCSUH` endpoint in SC-RunoffModelingServices

If "Lag Time Equation" is selected in the "Select Time of Concentration" dropdown
* Length (feet): StreamStats basin characteristic LFPLENGTH
[...]

PR #1401, opened 2022-08-08
CS-1385: SC Synthetic UH: Calculate parameters in user interface

KipSullivan:
Initial commit - there are still a couple of outstanding small issues to be addressed. Wanted to get your feedback on the current progress with #1385 and see if there is room for improvement/suggestions to improve/change it.

- Rearranged 'Select Time Of Concentration' to be toward the top, along with compacting the bottom half of the form (fields like 'Drainage Area' and 'Peak Rate Factor' are combined into a 2x3 grid
     - Still wondering if maybe the 'Time of Concentration, Tc' should be moved higher up the form as well.
newpagelayout

- 'Calculate Missing Parameters' button on the bottom of the page calculates most of the parameter fields automatically, using the endpoints listed in #1385
autocalculate

- More descriptive warning below the 'Calculate Missing Parameters' button to help user identify what needs to be filled out
calculationrequirements

Things yet to be done:
- If the drop downs are left blank, there isn't an orange outline indicating that they are blank, which are present for other text/number input fields. 
- Potentially move 'Time of Concentration, Tc' field higher up the form as well.
- More descriptive error handling if one of the parameters returned from the 'calculateMissingParametersSCSUH1' endpoint is missing or an unexpected value

### DIFF
--- a/src/Controllers/SCStormRunoffController.ts
+++ b/src/Controllers/SCStormRunoffController.ts
@@ -54,6 +54,7 @@ module StreamStats.Controllers {
         public canContinue: boolean;
         public ReportData: ISCStormRunoffReportable;
         public drainageArea: number;
+
         // SYNTHETIC TAB
         public drainageAreaSynthetic: number;
         public timeOfConcentrationMin: number;
@@ -63,8 +64,6 @@ module StreamStats.Controllers {
         public initialAbstraction: number;
         public lagTimeLength: number;
         public lagTimeSlope: number;
-        public syntheticParameters;
-        //
 
         public mainChannelLength: number;
         public mainChannelSlope: number;
@@ -87,7 +86,6 @@ module StreamStats.Controllers {
             }//end if           
         }
 
-        // 
         public AEPOptions = [{
             "name": "50%",
             "value": 50
@@ -954,7 +952,7 @@ module StreamStats.Controllers {
                 this.Execute(request).then(
                     (response: any) => {
                         if (response.data.parameters && response.data.parameters.length > 0) {
-                            //check each returned parameter for issues
+                            // check each returned parameter for issues
                             var paramErrors = false;
                             angular.forEach(response.data.parameters, (parameter) => {
                                 if (!parameter.hasOwnProperty('value') || parameter.value == -999) {
@@ -966,7 +964,7 @@ module StreamStats.Controllers {
                                     parameter.loaded = true;
                                 }
                             });
-                            //if there is an issue, pop open 
+                            // if there is an issue, pop open 
                             if (paramErrors) {
                                 this.toaster.pop('error', "Error", "Parameter failed to compute", 0);
                             }
@@ -976,14 +974,14 @@ module StreamStats.Controllers {
                                 if (param.code.toLowerCase() == 'lfplength' && this._selectedTimeOfConcentration?.value == 2 && !this.lagTimeSlope) this.lagTimeSlope =  param.value;
                             });
 
-                            //** */ Other SCStormModelingServices data
+                            // Other SCStormModelingServices data
                             var data = {};
                             var url = ""; 
                             var headers = {
                                 "Content-Type": "application/json",
                                 "X-warning": true
                             };
-                            let formmatedSegments = this.getFormattedFlowSegments();
+                            let formattedSegments = this.getFormattedFlowSegments();
                             // master data endpoing
                             data = {
                                 "lat": this.studyAreaService.selectedStudyArea.Pourpoint.Latitude,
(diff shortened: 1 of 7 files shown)


amedenblik (changes requested):
This is coming together nicely!

- [x] Can you edit the calculated values, so they don't have so many digits? There is an existing function `.toUSGSValue()` that will limit the values to the appropriate number of significant digits. 

image

- [x] Let's move "Select a CN Modification Method" below "Select a Rainfall Distribution Curve", since the CN Modification Method isn't one of the parameters that is calculated automatically.

- [x] I'm not sure the Rainfall Distribution Curve is being calculated correctly. For the point I tried (34.01380519814052, -80.99063158035278), the service returned `rainfall_distribution_curve_letter: "B"` but the drop-down showed "Type II" (should be "NOAA B").

- [x] If any values are manually entered, we don't want to overwrite those when we click "Calculate Missing Parameters". For example, if I enter Length and Slope for the Lag Time Equation, these values should not be overwritten when I click "Calculate Missing Parameters".

- [x] I think the "Channelized Flow - Storm Sewer Segments" are not getting translated correctly to the payload for the services. I entered 4 segments with pipe materials CMP, PVC, Concrete, and Steel, but the payload included pipe materials Aluminum, PVC, Concrete, and Concrete. 

- [x] We could rearrange the inputs like this, to better match how things are organized in the spreadsheet. I am open to other ideas, though! We may get additional feedback from SCDOT too.
image

- [x] "If the drop downs are left blank, there isn't an orange outline indicating that they are blank, which are present for other text/number input fields." -- if you can get this working, that would be cool! It's okay if not. 

- [x] "More descriptive error handling if one of the parameters returned from the 'calculateMissingParametersSCSUH' endpoint is missing or an unexpected value" this seems useful. 

Let us know if you could use any assistance or have further questions.

KipSullivan:
- Rearranged calculated inputs, while also limiting the number of significant digits of the calculations and not overriding any pre-existing values
dontoverwrite

- Adding yellow outline to empty drop down fields
warningondropdowns

- Fixed "Channelized Flow - Storm Sewer Segments" translation issue for endpoint calculation
- Fixed "Rainfall Distribution Curve" not being calculated correctly
- Moved "Select a CN Modification Method" below "Select a Rainfall Distribution Curve"
- More descriptive error handling for params that aren't calculated correctly in the "calculateMissingParametersSCSUH" endpoint

src/Controllers/SCStormRunoffController.ts
```diff
@@ -62,6 +63,9 @@ module StreamStats.Controllers {
         public initialAbstraction: number;
         public lagTimeLength: number;
         public lagTimeSlope: number;
+        public syntheticParameters;
+        //
```
harper-wavra: Cleanup unused comment

src/Controllers/SCStormRunoffController.ts
```diff
@@ -884,6 +900,179 @@ module StreamStats.Controllers {
...
+
+                            //** */ Other SCStormModelingServices data
+                            var data = {};
+                            var url = ""; 
+                            var headers = {
+                                "Content-Type": "application/json",
+                                "X-warning": true
+                            };
+                            let formmatedSegments = this.getFormattedFlowSegments();
+                            // master data endpoing
```
harper-wavra, line 985: spelling

src/Controllers/SCStormRunoffController.ts
```diff
@@ -884,6 +900,179 @@ module StreamStats.Controllers {
...
+                            if (paramErrors) {
+                                this.toaster.pop('error', "Error", "Parameter failed to compute", 0);
+                            }
+                            response.data.parameters.forEach(param => {
+                                if (param.code.toLowerCase() == 'drnarea' && !this.drainageAreaSynthetic) this.drainageAreaSynthetic = param.value;
+                                if (param.code.toLowerCase() == 'csl10_85fm' && this._selectedTimeOfConcentration?.value == 2 && !this.lagTimeLength) this.lagTimeLength = param.value;
+                                if (param.code.toLowerCase() == 'lfplength' && this._selectedTimeOfConcentration?.value == 2 && !this.lagTimeSlope) this.lagTimeSlope =  param.value;
+                            });
+
+                            //** */ Other SCStormModelingServices data
```
harper-wavra: comment is formatted a bit weird

harper-wavra (changes requested):
I think the calculate missing parameters button is working really well, and from my testing I think you were able to fix all the functionality notes that Andrea addressed. I just made a couple code clean-up comments. 

- [x] Andrea and I talked and we think we should move the CN modification Method (selected manually) above the rainfall distribution curve (can be calculated automatically). That way all the values that _have_ to be entered by the user at the top.

amedenblik (commented):
I don't have any additional comments beyond what Harper included in her review!

harper-wavra (approved):

KipSullivan:
Just cleaned up the code a bit following the suggestions from @harper-wavra - let me know if I missed anything!
- Moved 'CN Modification Method' drop down above the 'Rainfall Distribution Curve' drop down
- Removed unnecessary comments in code and fixed spelling on variables

harper-wavra:
Those changes all look good, thanks!

PR #1422, opened 2022-08-31
Fixes #1410

### LINKED ISSUE #1410: Update "Hydrologic Features" application
The latest Hydrologic Features updates (#1403) were made with the assumption that the report table was showing *all* NHD streams within the delineated basin. Instead, only the NHD streams that intersect the boundary of the delineated basin are shown. We need to make some updates to the report to indicate this. We also want to make some updates based on feedback from Jimmy.

- [x] Rename report section title: "NHD Features of Delineation Basin"
- [x] Rename table title: "NHD Streams Intersecting Basin Delineation Boundary"
- [x] Add description to the top of the report section, something like "This functionality attempts to find the stream name at the delineation point."
- [x] Add back disclaimer (from the modal) about differences in NHD vs str900 (the purpose of this functionality)
- [x] Add back Stream Name to the top of the report. Include a "?" (or similar) icon to point the user to the bottom section, which includes more information
- [x] Allow user to select stream with radio button
- [x] Add option for no stream/unknown/Other/manual entry/etc.
- [x] Remove "NHD Hydrologic Features" checkbox on sidebar-- no option to toggle this off

harper-wavra:
- Allows users to select different stream via radio buttons in report
- Removed "NHD Hydrologic Features" checkbox on sidebar
- Added back stream name to top of report with '?' icon that includes additional text
- Add back disclaimer from modal and a description to the top of the report section. Let me know if different text should be used. 
- Rename table title: "NHD Streams Intersecting Basin Delineation Boundary" & Rename report section title: "NHD Features of Delineation Basin"


amedenblik (approved):
Looks good! I made some small text changes and fixed a typo that I wrote in the original issue.

PR #1425, opened 2022-09-15
Fixes #1388

### LINKED ISSUE #1388: SC Synthetic UH: Allow report to be printable
See parent issue #1344 for more details
Complete #1386 first

- [x] Ensure that the modal can be printed and includes all information relevant to the SC Synthetic Unit Hydrograph Method, when the user clicks the "Print" button.`

- [x] Ensure modal values can be exported and includes all information relevant to the SC Synthetic Unit Hydrograph Method, when user clicks the "Download CSV."

harper-wavra:
- Enabled print button for the Synthetic Unit Hydrograph.
- Updated changelog for Synthetic Unit Hydrograph.
- Rebuilt.


amedenblik (changes requested):
I think we may want to change how the D-Hour Storm Hydrograph Ordinates formatting works, because there is currently no way to view the full table by printing the report: 

By default, the table doesn't appear at all: 
image

If you expand the section, you can only see a small part of the table:
image

I propose that when the "D-Hour Storm Hydrograph Ordinates" section is expanded, it shows the full table instead of a short scrollable table. Open to other ideas, though!

amedenblik (approved):

PR #1433, opened 2022-09-27
Fixes #1392

### LINKED ISSUE #1392: SC Synthetic UH: Display warning messages
See parent issue #1344 for more details
Complete https://github.com/USGS-WiM/SC-RunoffModelingServices/issues/66 first

At the top of the output section, add warnings in a yellow warning box.

harper-wavra:
- Added notes to bottom of SC Unit Hydrograph results
- Fixed bug when downloading csv and D-Hour results are not open


amedenblik (changes requested):
Works well! I would recommend making each of the notes an item <li> in an ordered list <ol> for formatting purposes. I would also make the text smaller, since it is currently larger than the header ("Notes on Time of Concentration:").

harper-wavra:
I was able to edit the text size a bit, let me know if that is better. 
I tried making them a list originally but the output to the csv was being very finicky with it as was the style of the modal once I added the `<ul>` and `<li>` tags.

amedenblik (changes requested):
Ah, I hadn't considered the difficulty with the `<ol>` and the CSV formatting! It is looking better. Do you think you can add a little space between the points? Maybe 2 `<br>` instead of 1?

amedenblik (approved):

PR #1451, opened 2022-10-21
Fixes #1394

### LINKED ISSUE #1394: Fix FDCTM issues
In general:

- [ ] Results in Iowa (and maybe other regions?) do not appear sometimes
image

Exceedance Probabilities graph:
- [x] The y-axis tick labels sometimes do not show up completely or overlap the y-axis label
image

Estimated Flows graph:
- [x] The y-axis tick labels and the y-axis label do not show up. Probably just need to change the left margin 
image

- [x] Change the series labels to be "Observed" and "Estimated"
image

- [x] Both series are the same color (dark blue) instead of dark blue and light blue 

Estimated Flows table:

- [x] Change observed flow column to be "Observed ([gage number])"

image

harper-wavra:
- Changed observed flow column in report to "Observed ([gage number])"
- Change the series labels for estimated flow plot to be "Observed" and "Estimated"
- Made both plots margins the same and increased left margin to 100. 

Other items from #1394 
- Not sure when this was fixed but I am seeing observed and estimated to be two different colors of blue now, so no action was needed. 
- Added bug about IA results not appearing (which is related to small basins having an RSD basin characteristic value of undefined) to this issue, https://github.com/USGS-WiM/StreamStats/issues/1261, since I think they will all be related.


src/Views/reportview.html
```diff
@@ -241,7 +241,7 @@ <h3 class="reportChartHeader">Estimated Flows</h3>
                                                             <tr>
                                                                 <th>Date</th>
                                                                 <th>Estimated</th>
-                                                                <th>{{r.referanceGage.name}}</th>
+                                                                <th>Observed ([{{r.referanceGage.stationID}}])</th>
```
amedenblik: Remove the brackets so that it says "Observed (05484500)". Sorry for the lack of clarity in the issue!

amedenblik (changes requested):

amedenblik (approved):

Issue #1201, opened 2022-02-23
Add IN Coordinated Reach Functionality


amedenblik:
- [x] Add the fixed Coordinated Reach code from #1178 
- [x] Add an IN layer called "Reaches": https://gis.streamstats.usgs.gov/arcgis/rest/services/coordinatedreaches/in/MapServer/
- [x] Include a Coordinated Reach flowstat report that includes (at minimum) these fields: StreamID/ID, Stream Name, and Basin Name(optional).

harper-wavra:
@USGSPMM3 The first two bullet points should be easy enough to add in. As for the 3rd item in the list, how would you like the Coordinate Reach flowstat report to be formatted? It sounds like he wants StreamID and Stream Name added for sure and Basin Name is optional. The [report.pdf](https://github.com/USGS-WiM/StreamStats/files/8143575/report.pdf)
currently only shows the basin name as far as I can tell. Where would you like us to add StreamID and Stream Name to the report?

Also do you have a preference for how the coordinate reach layer is displayed? I think Ryan might have mentioned making it yellow so it sticks out from the streamgrid layer.

USGSPMM3:
@harper-wavra -sorry for not seeing this:

For the Layer I think yellow is fine. I was also contemplating a thicker blue line so it looks like river reaches but is more prominent. Your call.

In the report we show the basin name in the parenthesis, can we just add the StreamID and name? Or replace the basin with stream ID and name? I don't know if Moon needs all 3?

PR #1464, opened 2022-11-08
Indiana Coordinated Discharge - add begin/end drainage area

amedenblik:
I was looking at old emails and saw one from Moon Kim on 2/18/22- "The CDS flowstat report is missing some essential fields from "eqWithStrID" table (please see attachments). I was able to compare a old report to the newer one (both attached) and newer one is missing Stream_ID/ID, Basin Name (currently labeled incorrectly as Coordinated Reach), Stream Name, Begin Drainage Area, and End Drainage Area. These field info is essential in both checking and confirming that the report matches the user selected stream reach. Please add these fields to the CDS flowstat report."

We had already added Stream_ID, Basin Name, and Stream Name. I added Begin Drainage Area and End Drainage Area. Please let me know if we decided against this at some point. I'm not sure why they weren't already added.

Attachments:

- This is a now-outdated StreamStats report: [StreamStatsUSGS 04101225 ST. JOSEPH RIVER AT SOUTH BEND, IN2.pdf](https://github.com/USGS-WiM/StreamStats/files/9956890/StreamStatsUSGS.04101225.ST.JOSEPH.RIVER.AT.SOUTH.BEND.IN2.pdf)

- Remove the .txt at the end of the file name to view the old StreamStats report: [fs03376500_cr.htm.txt](https://github.com/USGS-WiM/StreamStats/files/9956892/fs03376500_cr.htm.txt)

When pull request is complete and functionality has been pushed to Dev, we should follow up on Moon's email (subject line "[EXTERNAL] Re: unable to get correct coordinated discharge stats")


harper-wavra:
Seems like there was a lot and back and forth in that ticket and it probably just got lost. However, I think we created this issue (https://github.com/USGS-WiM/StreamStats/issues/1201) from the below email Moon sent in which he only asked for those three. I don't see a problem in adding these additional fields though.   
image

harper-wavra (approved):
Works well. May just want to add to the changelog, then looks good to merge.

amedenblik:
Pushed to Test and emailed Moon.

PR #1483, opened 2022-12-10
Fixes #1479 - Fix Print for Gage Page

### LINKED ISSUE #1479: Fix Print for Gage Page
https://streamstats.freshdesk.com/a/tickets/117870

amedenblik:
Closes #1479 

- Fixed the styling so the gage page can be printed with all data visible
- Added ability to download gage page as a CSV
- Also fixed the disclaimers that appear at the bottom of the main StreamStats report so that they are no longer delimited by commas in the text when downloaded as a CSV


harper-wavra (approved):
Works great!

PR #1496, opened 2023-01-05
Fixes #1424 - Annual Flow Plot with Flood-Frequency Statistics

### LINKED ISSUE #1424: Annual Flow Plot with Flood-Frequency Statistics
**General Information**: We would like to add a new modal popup for each stream gage that includes a plot of the annual peak flows with horizonal lines for the flood-frequency statistics. The idea of this plot comes from https://www.weather.gov/source/aprfc/gageAnalysis.html?site=12340000, [second plot on the left](https://user-images.githubusercontent.com/59063181/190000736-a8414fd3-1c54-438c-be04-bb0ff9a8483a.png) and [this image](https://user-images.githubusercontent.com/59063181/190183917-45afcf1c-5a11-4d08-ad03-9943c99d74f8.png). 

**Overview of Functionality**: When you zoom in past a level 7 on StreamStats the gages will start to appear as colored triangles. When you click on one of those icons a gage popup will appear. Inside that popup is some general information about the gage as well as a couple of links, including links to NWIS and a link to our StreamStats Gage Page modal. We would like to add another link, that follows in the same way the StreamStats Gage Page link works to open a new modal. Inside that modal is where we would like to display this plot. The plot should display the annual peaks, retrieved from NWIS, and the x-percent AEP flood statistics as horizontal lines retrieved from our own GageStatsServices.

**Steps to Complete**:
1. Create a link which opens a new modal in the gage popup. You will have to create a new view and controller for this step. See the [Gagepage.html](https://github.com/USGS-WiM/StreamStats/blob/dev/src/Views/gagepage.html) and [GagePageController.ts](https://github.com/USGS-WiM/StreamStats/blob/dev/src/Controllers/GagePageController.ts) as examples.
2. Decide which plotting library to use. In order to create the plot, we will need to use a plotting library. The developer for the weather.gov plots used [HighCharts](https://www.highcharts.com/blog/tutorials/194-using-highcharts-with-angular-js/) so it could be helpful to use the same. Within StreamStats we have also used the library [NVD3](https://nvd3.org/), so that could be another option. If you search for 'nvd3' in the code, you can see where we have used it already. The future plan is to recreate most of the plots you see on that weather.gov page, so using HighCharts might be best in the long run. 
3. Pull annual peaks values from NWIS to begin to create a plot in the modal. The nwis call you can use is: https://nwis.waterdata.usgs.gov/usa/nwis/peak/?format=rdb&site_no=02173500&startDT=2021-04-29&endDT=2022-12-30. Date can be on the x axis and Discharge can be on the y axis, then each annual peak value will be a point on the plot. 
4. Display horizonal lines for the flood-frequency statistics. These values come from our gage statistics, we want to be able to plot all of the x-percent AEP flood values that a gage has.
image
The above image shows all of the x-percent AEP flood values that a gage _could_ have, sometimes they will only have some of these and sometimes they will have none. To get all the statistics for a certain gage you will use the service call: https://test.streamstats.usgs.gov/gagestatsservices/stations/02173500. For a better view of what this json looks like you can use our [service documentation ](https://test.streamstats.usgs.gov/docs/gagestatsservices/#/Stations/GET/Station) or just find a json viewer online (https://jsonformatter.org/json-pretty-print). After you make that call and get the returned object you will need to sort through the `statistics` array and get the `value` of those with `regressionTypeID` equal to the IDs in the image above. So look for IDs equal to 9, 852, 8, 4, 7, 3, 6, 1, 501, 5, 2, 500, 851, 1438, then plots those values as horizontal lines in the plot.
5. Plot the daily values as a line (this is the blue line from the NWS Gage Analysis Plot). The call to NWIS for daily values is: https://nwis.waterservices.usgs.gov/nwis/iv/?format=json&sites=02173500&parameterCd=00060&startDT=2022-04-29&endDT=2022-12-30. 
6. Add inputs so the user can enter a start and end date. Have two input boxes that allow the users to select a start date and an end date, then once the user clicks a button the plot should update to only show the values between those two dates. You should be able to configure the two NWIS calls with your start and end dates. 

**Functionality we also want from the NWS Gage Analysis Plot**:
- [ ] Show annual peaks and horizonal flood statistic lines with same colors - I think we need more information on this
- [x] Zoom into graph

**Tips:**
- If you go to https://www.weather.gov/source/aprfc/gageAnalysis.html?site=12340000, right click "View Source", you will be able to view the source code for that webpage along with all of its plots. The developer who wrote this page said we could look and use the code. The plot we want to recreate has an id of "dischargegraph." A couple of the functions and sections of code that are used to create that plot and that could be helpful to look at are
    - Lines 97-165 - HTML code used to display the plot
    - Line 4407 - 4431 - LoadQ, used to get new data after new dates and the 'Load Primary Data' button is clicked.
    - Lines 4520-4579 - Gets the nwis data for plotting
- Thresholds also does something similar. https://test.wim.usgs.gov/thresholds/#/ zoom into somewhere and turn on the streamgages layer and click on a gage. It will show a plot of the daily values. The code for Thresholds is here, https://github.com/USGS-WiM/Thresholds.
- Make new branches based on the SC-Runoff branch for this issue. Pull requests should merge into the SC-Runoff branch. We will need to manually close issues after merge requests are completed, since we are not merging into the default branch
- Old issue https://github.com/USGS-WiM/StreamStats/issues/1371 (has some helpful images)

maggiejaenicke:
Closes #1424


src/Controllers/GagePageController.ts
```diff
@@ -536,13 +581,300 @@ module StreamStats.Controllers {
...
+                }, (error) => {
+                    //sm when error
+                }).finally(() => {
+                });
+        }
+
+        //Get peak values from NWIS
+        public getPeakInfo() {
+            const url = 'https://nwis.waterdata.usgs.gov/usa/nwis/peak/?format=rdb&site_no=' + this.gage.code
+            console.log('GetPeakURL', url)
```
amedenblik: Comment out or delete all console.log statements

src/Controllers/GagePageController.ts
```diff
@@ -536,13 +581,300 @@ module StreamStats.Controllers {
...
+            this.gage = new GageInfo(this.modalService.modalOptions.siteid);
+            //below: how to get the data from API and setting it in controller
+            const url = configuration.baseurls.GageStatsServices + configuration.queryparams.GageStatsServicesStations + this.gage.code;
+            const request: WiM.Services.Helpers.RequestInfo = new WiM.Services.Helpers.RequestInfo(url, true, WiM.Services.Helpers.methodType.GET, 'json');
+
+            this.Execute(request).then(
+                (response: any) => {
+                    this.gage = response.data;
+                    this.getPeakInfo();
+                    this.getFloodFreq();
```
amedenblik: `this.getFloodFreq()` is called on this line and also on line 646. Same with `this.getDailyFlow()`-- it is called here and on line 672. I think these functions are being run twice, when they only needs to run once? If so, you probably want to omit these two lines (598 and 599).

src/Controllers/GagePageController.ts
```diff
@@ -536,13 +581,300 @@ module StreamStats.Controllers {
...
+                    zooming: {
+                        type: 'xy'
+                    }
+                },
+                title: {
+                    text: 'Annual Peak Streamflow',
+                    align: 'center'
+                },
+                subtitle: {
+                    text: 'Click and drag in the plot area to zoom in<br>AEP: Annual Exceedance Probability',
```
amedenblik: Good idea to describe AEP. Let's say "AEP = Annual Exceedance Probability" instead of the colon.

src/Controllers/GagePageController.ts
```diff
@@ -536,13 +581,300 @@ module StreamStats.Controllers {
...
+                    type    : 'scatter',
+                    color   : 'black',
+                    data    : this.formattedPeakDates,
+                    marker: {
+                        symbol: 'circle',
+                        radius: 3
+                    }
+                },
+                {
+                    name    : 'Annual Peak Streamflow (Date Estimated)',
```
amedenblik, line 812: This data series is showing up in the legend, even if there are no Annual Peak Streamflows that have an estimated value. Is it possible to not include this series, if there are no applicable values? You could check to see if the length of `estPeakValues` is greater than 0 to determine the true/false logic of whether or not to include the series.

src/Controllers/GagePageController.ts
```diff
@@ -536,13 +581,300 @@ module StreamStats.Controllers {
...
+                    tooltip: {
+                        headerFormat:'<b>Peak Annual Flow</b>',
```
amedenblik: Let's call this "Annual Peak Streamflow" so that it matches what is listed in the legend.

src/Views/gagepage.html
```diff
@@ -195,10 +195,28 @@ <h4 class="modal-title">StreamStats Gage Page</h4>
...
             </div>
 
-            <!-- Gage Analysis Plots Tab-->
+            <!-- Gage Analysis Plots Tab -->
             <div ng-if="vm.SelectedTab == 2" style="position: relative;">
                 <div style="padding-top: 10px;">
-                    Gage Analysis Plots will go here
+                    <div class="panel panel-default">
+                        <div class="panel-heading">Gage Information</div>
+                            <table class="table">
```
amedenblik: For this table, see if you can copy the styling (the column widths and the the background colors of the rows) that is seen on the "Gage Information" tab:
image

src/Views/gagepage.html
```diff
@@ -195,10 +195,28 @@ <h4 class="modal-title">StreamStats Gage Page</h4>
...
+                                <tbody>
+                                    <tr>
+                                        <th scope="row">USGS Station ID: </th>
```
amedenblik: "USGS Station Number" instead of "USGS Station ID:"

src/Views/gagepage.html
```diff
@@ -195,10 +195,28 @@ <h4 class="modal-title">StreamStats Gage Page</h4>
...
+                                            <td>{{vm.gage.code}}</td>
```
amedenblik: Make the station code a hyperlink, as seen in the "Gage Information" tab

src/Views/gagepage.html
```diff
@@ -195,10 +195,28 @@ <h4 class="modal-title">StreamStats Gage Page</h4>
...
+                                    </tr>
+                                    <tr>
+                                        <th scope="row">Station Name:</th>
```
amedenblik: Remove colon: "Station Name"

src/Views/gagepage.html
```diff
@@ -195,10 +195,28 @@ <h4 class="modal-title">StreamStats Gage Page</h4>
...
+                                            <td>{{vm.gage.name}}</td>
+                                    </tr>
+                                </tbody>
+                            </table>
+                    </div>
+                    <div class="panel panel-default">
+                        <!-- <div class="panel-heading">Annual Peak Streamflow</div> -->
+                        <highchart id="chart1" config="vm.chartConfig"></highchart>
```
amedenblik, line 227: Is it possible to horizontally center the chart?

amedenblik (changes requested):
Awesome work! I'm excited to have our first graph in the new Gage Analysis Plots tab! 🤩

Please see the review comments in the code. 

Additionally, I noticed that there are 2 sets of lines for some gages (ex. 12340000):
image

This is happening because there are 2 sets of AEP statistics, from 2 separate publications, and all are marked as "isPreferred". I have asked Pete for guidance on how we should deal with this, so stay tuned on that.

Update from Pete: there should only be one set of "isPreferred" statistics in the database. He will work on the database and we will stay tuned. Let's not worry about this problem for this PR.

src/Controllers/MapController.ts
```diff
@@ -1451,7 +1450,7 @@ module StreamStats.Controllers {
                     },
                     onEachFeature: function (feature, layer) {
                         var siteNo = feature.properties['Code'];
-                        var NWISpage = 'https://waterdata.usgs.gov/monitoring-location/' + siteNo;
+                        var PeakNWISPage = 'http://nwis.waterdata.usgs.gov/usa/nwis/peak/?site_no=' + siteNo;
```
harper-wavra: This variable is not used and can be removed.

src/Controllers/GagePageController.ts
```diff
@@ -536,13 +580,285 @@ module StreamStats.Controllers {
...
+                    type: 'datetime',
+                    title: {
+                        text: 'Date'
+                    },
+                },
+                yAxis: {
+                    title: {
+                        text: 'Discharge (Q), in ft³/s'
+                    },
+                    plotLines: []
```
harper-wavra, line 754: Is this variable needed? It is throwing an error in the code. 

`Type '[]' is not assignable to type '[{ value: number; color: string; width: number; zIndex: number; label: { text: string; }; }]'. Source has 0 element(s) but target requires 1.`

src/Controllers/GagePageController.ts
```diff
@@ -536,13 +580,285 @@ module StreamStats.Controllers {
...
+                    this.dailyFlow = dailyValues
+                    this.formatData();
+                }); 
+            }
+
+        //Get data into (x, y) format and convert to dates in order to add it to the plot
+        public formatData(): void {
+            if (this.peakDates) {
+                this.formattedPeakDates = [];
+                this.peakDates.forEach(test => {
```
harper-wavra: Could we use `peakValue` instead of `test`? Having descriptive variables names will help us if we come back to this section of code later. 

I think there are some other spots that use `test` as a variable name that could also be changed.

src/Controllers/GagePageController.ts
```diff
@@ -536,13 +580,285 @@ module StreamStats.Controllers {
...
+
+        // Gage Analysis Plots methods
+
+        public getGagePlots() {
+            this.getPeakInfo(); // Annual Peak Streamflow plot
+        }
+
+        //Get peak values from NWIS
+        public getPeakInfo() {
+            const url = 'https://nwis.waterdata.usgs.gov/usa/nwis/peak/?format=rdb&site_no=' + this.gage.code
```
harper-wavra, line 581: I wonder if there is a way to add the start and end dates to this request. I looked for a bit but wasn't able to find an obvious way. I sent an email asking.  

If there is not an easy way, we will need a way to filter these by dates when we add in the user option to select a period of record. (but we can always figure that out in a new issue)

src/Controllers/GagePageController.ts
```diff
@@ -536,13 +580,285 @@ module StreamStats.Controllers {
...
+                        };
+                    } while (data.length > 0);
+                    const filteredArray = peakValues.filter(item => {
+                        return (item.peak_dt[8] + item.peak_dt[9] !== '00' || item.peak_dt[8] + item.peak_dt[9] !== '00') //filtering out invalid dates from main array
+                    });
+                    this.peakDates = filteredArray;
+                    this.estPeakDates = estPeakValues;
+                }, (error) => {
+                }).finally(() => {
+                    this.getFloodFreq()
```
harper-wavra: We try to finish each line of code with a `;`. It doesn't always happen but it's a good habit try to get into.

src/Controllers/GagePageController.ts
```diff
@@ -136,9 +144,22 @@ module StreamStats.Controllers {
         description: string;
     }
 
+    // Gage Analysis Plots classes
+    class peakValues {
```
harper-wavra: I don't think this class is being used (though I could be wrong). I think you set this again in line 605 & used that variable. If it's not being used it can be removed.

src/Controllers/GagePageController.ts
```diff
@@ -536,13 +580,285 @@ module StreamStats.Controllers {
...
+        public getPeakInfo() {
+            const url = 'https://nwis.waterdata.usgs.gov/usa/nwis/peak/?format=rdb&site_no=' + this.gage.code
+            console.log('GetPeakURL', url)
+            const request: WiM.Services.Helpers.RequestInfo = new WiM.Services.Helpers.RequestInfo(url, true, WiM.Services.Helpers.methodType.GET, 'json');
+            this.Execute(request).then(
+                (response: any) => {
+                    const peakValues = [];
+                    const estPeakValues = []; //dates that include a '-00'
+                    const data = response.data.split('\n').filter(r => { return (!r.startsWith("#") && r != "") });
+                    const headers:Array<string> = data.shift().split('\t');
```
harper-wavra: I don't think `headers` is being used anywhere, if that is true it can be removed.

harper-wavra (changes requested):
Yay this looks great! 🎊 

- [x] Not a huge deal but I wonder if there isn't a point for one of the legend items, we remove it from the legend. For example, on station 06342500, there is not an estimated point. Could we remove that from the legend? [example](https://user-images.githubusercontent.com/59063181/211035918-f911301d-bfaf-48ab-8a1a-888704585373.png)
- [x] We might have talked about this and I just forgot - Is there a way to add a tool tip to the AEP lines? It could be helpful to be able to hover over those lines and see their values. **- Issue made: https://github.com/USGS-WiM/StreamStats/issues/1498**

If any of the changes we have mentioned are looking to be difficult or time consuming - please feel free to create new & separate issues for them.

src/Controllers/GagePageController.ts
maggiejaenicke: That's the container that the AEP data is getting pushed into to create the plotLines. I do need it there to get the plotLines to work. No error in the console, but I'm not sure if there's a way to write it where VS Code doesn't throw an error. Open to ideas.

harper-wavra: No worries if it is needed!

src/Views/gagepage.html
maggiejaenicke: Copied the colors and styling. Seems like column widths are correlated with the length of the content in them.

src/Controllers/GagePageController.ts
maggiejaenicke: I think I'll add this to the new issue along with adding the user option for start and end dates.

src/Controllers/GagePageController.ts
```diff
@@ -536,13 +569,287 @@ module StreamStats.Controllers {
...
+        // Gage Analysis Plots methods
+
+        public getGagePlots() {
+            this.getPeakInfo(); // Annual Peak Streamflow plot
+        }
+
+        //Get peak values from NWIS
+        public getPeakInfo() {
+            const url = 'https://nwis.waterdata.usgs.gov/usa/nwis/peak/?format=rdb&site_no=' + this.gage.code
+            console.log('GetPeakURL', url)
```
amedenblik: Comment out or delete all `console.log` statements-- there are a couple still in here

amedenblik (changes requested):
The changes look great! I left one tiny comment.

amedenblik (approved):

src/Controllers/GagePageController.ts
harper-wavra: Issue made: https://github.com/USGS-WiM/StreamStats/issues/1498

harper-wavra (approved):
Looking good to me!

harper-wavra:
Do we want to set up a separate dev site for this branch so Pete and others (South Carolina?) can test it out?

amedenblik:
Yes, I will set up a separate dev site 👍

PR #1501, opened 2023-01-10
Fixes #1500

### LINKED ISSUE #1500: Gageplots loading issue
When the `gageplots` branch is deployed (https://dev.streamstats.usgs.gov/gageplots/), the plot does not load on the Gage Analysis Plots tab of the Gage Page:

image

There are numerous console errors:
image

This issue does not happen when running the `gageplots` branch locally, but I suspect that it occurs because the gage plot is being rendered before the data is ready. A first attempt at this issue would be initiating the gage plot *after* all dependent data has been loaded. We should also consider adding a spinning loader for the location where the plot should be.

harper-wavra:
Should be working on https://dev.streamstats.usgs.gov/gageplots/

- Added highcharts - see https://stackoverflow.com/questions/41244791/cannot-read-property-chart-of-undefined-highcharts-ng
- added spinner while we wait for data to be returned
- Gave default values to `plotLines` to remove errors inline code errors
- Was seeing errors with `this.formattedPeakDates.length > 0` & `this.formattedEstPeakDates.length > 0` when `formattedPeakDates` or `formattedEstPeakDates` where not defined. I change those two lines of code to match `showInLegend: this.formattedDailyFlow != undefined`.


src/Controllers/GagePageController.js
```diff
@@ -642,13 +643,14 @@ var StreamStats;
...
                             },
-                            showInLegend: this.formattedEstPeakDates.length > 0
+                            showInLegend: this.formattedEstPeakDates != undefined
                         }
                     ]
                 };
                 this.formattedFloodFreq.forEach(function (formattedFloodFreqItem) {
                     _this_1.chartConfig.yAxis.plotLines.push(formattedFloodFreqItem);
                 });
+                console.log(this.chartConfig);
```
maggiejaenicke: Will we need to clear out/comment console logs? Feel free to ignore this if not, but that was something brought up in my original PR.

src/Controllers/GagePageController.ts
```diff
@@ -835,7 +834,7 @@ module StreamStats.Controllers {
                         symbol: 'square',
                         radius: 3
                     },
-                    showInLegend: this.formattedEstPeakDates.length > 0
+                    showInLegend: this.formattedEstPeakDates != undefined
```
maggiejaenicke: With this change I'm now seeing that Estimated Peak Streamflow is showing up in the legend but have no data points (ex: 02173351) on sites that **do have** regular peak values. The showInLegend line is working correctly at sites that have **neither** peak streamflow values or estimated peak streamflow values (ex: 02156450). 

This seems to be happening because when there is normal peak date data (but no invalid dates), it results in an array with a length of 0 for the estimated peaks (not undefined), so it shows in the legend. When there is no normal peak date data present, both the Peak Dates and Estimated Peak dates are undefined, which is what is currently working. 

I'm not sure the best solution here, since neither _.length >0_ or _!= undefined_ will cover both scenarios, but maybe there's another way to write this that will handle both. 

I think _!= undefined_ works well for the other two series.

maggiejaenicke (changes requested):
Looks great! I'm glad you got the chart to load!

The only other thing I noticed was that charts with no data no longer have the 'No data to display message'. Instead, it's an empty plot. I'm not sure if that happened in these changes or earlier, maybe something we should revisit. 

empty_plot

src/Controllers/GagePageController.js
amedenblik:
```suggestion
```

src/Controllers/GagePageController.ts
amedenblik: To fix this, I recommend changing this (lines 676-681):
```
            if (this.estPeakDates) {
                this.formattedEstPeakDates = [];
                this.estPeakDates.forEach(estPeakObj => {
                    this.formattedEstPeakDates.push({x: new Date(estPeakObj.peak_dt), y: estPeakObj.peak_va})
                });
            }
```

to this:

```
            if (this.estPeakDates) {
                this.estPeakDates.forEach(estPeakObj => {
                    if (!this.formattedEstPeakDates) {
                        this.formattedEstPeakDates = [];
                    }
                    this.formattedEstPeakDates.push({x: new Date(estPeakObj.peak_dt), y: estPeakObj.peak_va})
                });
            }
```

amedenblik (changes requested):

src/Controllers/GagePageController.ts
harper-wavra: It doesn't go into that loop unless there is `estPeakDates`, which in the case of [06342500] there isn't. I'll look into it more.

harper-wavra:
I think I got it working, I just initialized all the formatted arrays at the beginning and then to set the legend I am always looking for length greater than one. I tested a few sites and things seemed to be working but let me know if you find any that are not.

amedenblik (approved):

maggiejaenicke (approved):
Looks good!

harper-wavra:
I can update the dev site.

harper-wavra:
The dev site is now updated.

PR #1515, opened 2023-01-26
Fixes #1498 Additional Functionality for Annual Flow Plot with Flood-Frequency Statistics

### LINKED ISSUE #1498: Additional Functionality for Annual Flow Plot with Flood-Frequency Statistics
**Tips:**
- Make new branches based on the gageplots branch
- See issue https://github.com/USGS-WiM/StreamStats/issues/1424 for details on the creation of these plots.
- See comments in pull request #1496 

**Functionalities to add:**

- [x] #1514

- [x] **Add inputs so the user can enter a start and end date.** Have two input boxes that allow the users to select a start date and an end date, then once the user clicks a button the plot should update to only show the values between those two dates. You should be able to configure the two NWIS calls with your start and end dates. For an example, see the [weather.gov page](https://www.weather.gov/source/aprfc/gageAnalysis.html?site=12340000), right click "View page source", and search for "dischargeStart" and "dischargeEnd". 

- [x] **Add a button to AEP PlotLines turn AEP PlotLines off and on.** Some charts have AEP lines that are very close together (ex: 06177720), which appear very cluttered. The user can zoom in to get a better view of the lines, but it might be useful to have a functionality to turn them off and on so that the user can view the entire plot without it being obscured by the lines. Something like [this](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/members/axis-addplotband/) might work. [Highcharts site](https://www.highcharts.com/docs/chart-concepts/plot-bands-and-plot-lines) documentation on adding and removing -- see "dynamically update."

- [x] Change the default zoom to "all" (part of the rangeSelector). Because the daily values are used for the navigator (timeline slider), the default zoom includes just the daily values. This usually isn't a problem and all the data are visible, but this plot (02173500) is an example of where you miss data on the initial load. 

- [x] Add the ability to view the plot in log scale. A checkbox would probably be ideal. This functionality is available in the [weather.gov plot.](https://www.weather.gov/source/aprfc/gageAnalysis.html?site=1234000). 

- [x] Add these statistic IDs: 2311, 2312, 2313, 2314, 2315, 2316, 2317, 2318

maggiejaenicke:
Let me know what you think about having all in data in the navigator (timeline slider). That was the only way I could make sure all data were shown on initial load. 

Closes #1498.


src/Controllers/GagePageController.ts
```diff
@@ -838,6 +876,31 @@ module StreamStats.Controllers {
                 this.chartConfig.yAxis.plotLines.push(formattedFloodFreqItem)
             });
         }
+
+        //checkbox for turning plotLines on and off
+        public plotlines = true;
+            public removePlotLines () {
```
amedenblik: Consider renaming this function to `togglePlotLines`, since it doesn't just remove the lines.

src/index.html
```diff
@@ -78,9 +78,14 @@
...
-    <script src="http://code.highcharts.com/highcharts.src.js"></script>
+    <!-- <script src="http://code.highcharts.com/highcharts.src.js"></script> -->
+    <script src="https://code.highcharts.com/stock/modules/stock.js"></script>
+    <!-- <script src="https://code.highcharts.com/stock/highstock.js"></script> -->
+    <script src="https://code.highcharts.com/stock/modules/data.js"></script>
+    <script src="https://code.highcharts.com/stock/modules/exporting.js"></script>
+    <script src="https://code.highcharts.com/stock/modules/accessibility.js"></script>
     <script src="https://code.highcharts.com/modules/no-data-to-display.js"></script>
-    <script src="http://code.highcharts.com/stock/highstock.src.js"></script>
+    <!-- <script src="http://code.highcharts.com/stock/highstock.src.js"></script> -->
```
amedenblik: Removed lines that are commented out

src/Controllers/GagePageController.ts
```diff
@@ -838,6 +876,31 @@ module StreamStats.Controllers {
...
+                });
+            }
+                else {
+                chart.yAxis[0].removePlotLine('plotlines'); // all plot lines have id: 'plotlines'
+                }
+            };
+
+        //checkbox for change linear to log scale
+        public logScale = false; 
+        public logToLinear () {
```
amedenblik: Consider renaming this function to something like `toggleLogLinear`, since it doesn't just change it from to log to linear

src/Controllers/GagePageController.ts
```diff
@@ -190,11 +190,12 @@ module StreamStats.Controllers {
         public peakValuePlot: any;
         public annualFlowPlot: any;
         public peakValues: any;
+        public removeAEPLines: boolean;
```
amedenblik: Can this variable be removed? I don't think it's used elsewhere.

amedenblik (changes requested):
Things are working well!

- [x] Unrelated to this issue: for gage 02169630 I'm seeing "Annual Peak Streamflow (Date Estimated)" in the legend, but I'm not seeing red dots on the plot: 
image

- [x] The Annual Peak Streamflow values are appearing in the navigator as a black line, rather than black dots. Is there a way to change this to dots? The line is a little misleading/distracting.
image

If the goal is to try to show the full extent of all data from the beginning, an alternative way to do this might be: figure out what is the earliest date among all the available data (Daily Streamflow, Annual Peak Streamflow, and Annual Peak Streamflow (Date Estimated)). Then set that to the min x-axis value. I think that would work?

harper-wavra (changes requested):
These changes are looking good!

- I don't think I am seeing the AEP lines on some plots. For example, 06471200 has a couple AEP flood statistics but I am not seeing those lines.

maggiejaenicke:
Thanks for both of your comments--I believe I fixed everything. To Harper's comment about seeing not seeing AEP lines on 06471200, I looked through the data and it looks like all the regressionTypeIDs we're looking for are marked "false" on isPreffered. If I'm missing something--happy to make changes to that!

harper-wavra (approved):
I think this is looking good! It will be nice to get this on the test stie so Pete and the SC cooperators can see.

amedenblik (approved):
Looks good! I can rebuild and update on https://dev.streamstats.usgs.gov/gageplots/

PR #1541, opened 2023-02-09
Fixes #1457: Add Select State/Region Dropdown

### LINKED ISSUE #1457: BP: Add "Select State/Region" dropdown
Part of #1455 

- [x] Create a dropdown that says "Select State / Region:"
- [x] Make a service call to https://streamstats.usgs.gov/nssservices/regions to retrieve the list of all the Regions
- [x] Populate the dropdown with the "name" of each Region 
- [x] Also create a checkbox that says "Basin Delineation", which gets forcibly selected when a State/Region is selected

Note: we may need to refine this, as there are some regions like "Undefined" that we may not want to include.

laws-spatial:
Closes #1457

### DIFF
--- a/src/Services/nssService.ts
+++ b/src/Services/nssService.ts
@@ -159,7 +159,6 @@ module StreamStats.Services {
         public isDone: boolean;
         public reportGenerated: boolean;
         private modalService: Services.IModalService;
-        public regionList: any;
 
         //Constructor
         //-+-+-+-+-+-+-+-+-+-+-+-
--- a/src/Services/nssService.ts
+++ b/src/Services/nssService.ts
@@ -41,7 +41,7 @@ module StreamStats.Services {
         queriedRegions: boolean;
         getflattenNSSTable(name: string): Array<INSSResultTable>
         reportGenerated: boolean;  
-        getRegionList();
+        getRegionList(): ng.IPromise<any>;
     }
     export interface IStatisticsGroup {
         id: string;
@@ -791,22 +791,29 @@ module StreamStats.Services {
         }
 
         // get region list form nssservices/regions
-        public getRegionList(): any {
+        // referenced getFreshDeskArticles from HelpControllers
+        // keys: use ng.Ipromise<any> and return this.Execute 
+        public getRegionList(): ng.IPromise<any> {
+
             var url = configuration.baseurls['NSS'] + configuration.queryparams['NSSRegions'];
             var request: WiM.Services.Helpers.RequestInfo = new WiM.Services.Helpers.RequestInfo(url, true);
-            this.Execute(request).then(
+            
+            return this.Execute(request).then(
                 (response: any) => {
-                    console.log("region list" + response);
-                    this.regionList = response.data;
-                    // return (this.regionList);
+                    // create array to return
+                    var regions = []; 
+
+                    // loop through response and add data to regions array
+                    response.data.forEach(function (element) {
+                        regions.push(element);
+                    });
+
+                    // console.log("regionList_nssServices", response.data);
+                    return regions
                 }, (error) => {
                     this.toaster.pop('error', "There was an HTTP error returning the regions list.", "Please retry", 0);
                 }).finally(() => {
-                }
-                );
-            // this.$http.get("https://streamstats.usgs.gov/nssservices/regions")
-            //     .then((data) => this.regionList = data);
-            // console.log(this.regionList);
+                });
         }
 
         //HelperMethods
--- a/src/Services/nssService.ts
+++ b/src/Services/nssService.ts
@@ -800,19 +800,17 @@ module StreamStats.Services {
             return this.Execute(request).then(
                 (response: any) => {
                     // create array to return
-                    var regionsRaw = []; 
+                    var regions = response.data;
 
-                    // loop through response and add data to regions array
-                    response.data.forEach(function (element) {
-                        regionsRaw.push(element);
-                    });
+                    /* uncomment to implement filter
+                    var regionsRaw = response.data; 
 
                     // add filter for unwanted values, additional values can be added with &&
                     function filterRegions(element, index, array) {
                         return element.name !== "Undefined";
                     }
 
-                    var regions = regionsRaw.filter(filterRegions);
+                    var regions = regionsRaw.filter(filterRegions); */
                     
                     // console.log("regionList_nssServices", regions);
 
--- a/src/Services/nssService.ts
+++ b/src/Services/nssService.ts
@@ -41,6 +41,7 @@ module StreamStats.Services {
         queriedRegions: boolean;
         getflattenNSSTable(name: string): Array<INSSResultTable>
         reportGenerated: boolean;  
+        getRegionList();
     }
     export interface IStatisticsGroup {
         id: string;
@@ -158,6 +159,7 @@ module StreamStats.Services {
         public isDone: boolean;
         public reportGenerated: boolean;
         private modalService: Services.IModalService;
+        public regionList: any;
 
         //Constructor
         //-+-+-+-+-+-+-+-+-+-+-+-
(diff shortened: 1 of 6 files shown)


src/Services/nssService.ts
```diff
@@ -158,6 +159,7 @@ module StreamStats.Services {
         public isDone: boolean;
         public reportGenerated: boolean;
         private modalService: Services.IModalService;
+        public regionList: any;
```
amedenblik:
```suggestion

```
I don't think this variable is used and can be removed

src/Services/nssService.ts
```diff
@@ -787,6 +789,41 @@ module StreamStats.Services {
...
+                    var regionsRaw = []; 
+
+                    // loop through response and add data to regions array
+                    response.data.forEach(function (element) {
+                        regionsRaw.push(element);
+                    });
+
+                    // add filter for unwanted values, additional values can be added with &&
+                    function filterRegions(element, index, array) {
+                        return element.name !== "Undefined";
```
amedenblik, line 810: Upon further inspection, I see there are several other regions we'll want to exclude. In the future, we will probably return a boolean variable to https://streamstats.usgs.gov/nssservices/regions to signify whether or not the region should be included in the batch processor list. We will make a new issue when that is ready.

src/Views/batchprocessor.html
```diff
@@ -15,7 +15,32 @@ <h4 class="modal-title">StreamStats Batch Processor</h4>
...
         <div id="submitBatch" ng-if="vm.selectedBatchProcessorTabName==='submitBatch'">
-            <p>This is the submit batch processor tab</p>
+            <div id="selectRegionID">
+                <label>Select Region / State</label>
+                <select ng-model="selectedRegion" ng-options="x.name for x in vm.regionList | orderBy: 'name'"></select>
+            </div>
+            <div>
+                <span>
+                    <label class="chx">
+                        <input id="cbDelineate", type="checkbox" ng-checked="selectedRegion">
```
amedenblik: Once the user selects a State/Region, the user should not be allowed to uncheck this box

src/Views/batchprocessor.html
```diff
@@ -15,7 +15,32 @@ <h4 class="modal-title">StreamStats Batch Processor</h4>
...
+                        <input type="checkbox">
+                        <span>&nbsp;&nbsp;Compute Basin Characteristics&nbsp;&nbsp;&nbsp;</span>
+                    </label>
+                </span>
+                
+                <span>
+                    <label class="chx">
+                        <input type="checkbox">
+                        <span>&nbsp;&nbsp;Compute Flow Statistics</span>
+                    </label>
```
amedenblik: When I check a checkbox, the checkbox moves upwards slightly. Is there a way to prevent this? 

Kind of hard to see in the image, but the first checkbox is slightly higher than the others:
image

src/Views/batchprocessor.html
```diff
@@ -15,7 +15,32 @@ <h4 class="modal-title">StreamStats Batch Processor</h4>
     <div style="padding-top: 10px;">
 
         <div id="submitBatch" ng-if="vm.selectedBatchProcessorTabName==='submitBatch'">
-            <p>This is the submit batch processor tab</p>
+            <div id="selectRegionID">
+                <label>Select Region / State</label>
```
amedenblik:
```suggestion
                <label>Select State / Region:</label>

```

amedenblik (changes requested):
Good work!

src/Services/nssService.ts
harper-wavra: Maybe we get rid of this filterRegions for now? Then work on the filtering of which regions should be displayed in that new issue when ready.

harper-wavra:
~~FYI - we just turned off the test server for production, so if you are working on this PR this morning you might be seeing a error with the https://test.streamstats.usgs.gov/nssservices/regions/. If you change to https://streamstats.usgs.gov/nssservices/regions/ it should work.~~ should be back on

src/Services/nssService.ts
```diff
@@ -787,6 +789,41 @@ module StreamStats.Services {
...
+            
+            return this.Execute(request).then(
+                (response: any) => {
+                    // create array to return
+                    var regionsRaw = []; 
+
+                    // loop through response and add data to regions array
+                    response.data.forEach(function (element) {
+                        regionsRaw.push(element);
+                    });
```
harper-wavra: You should be able to change these lines of code to:
`var regionsRaw = response.data; `
which should be more efficient than a loop.

harper-wavra (changes requested):
Looking good!

src/Services/nssService.ts
laws-spatial: I will comment out the filter until we implement the boolean variable to the nssservices, just in case we decide to use it in the future issue.

src/Views/batchprocessor.html
laws-spatial: fixed with ```onclick="return false;"

src/Views/batchprocessor.html
```diff
@@ -15,7 +15,24 @@ <h4 class="modal-title">StreamStats Batch Processor</h4>
     <div style="padding-top: 10px;">
 
         <div id="submitBatch" ng-if="vm.selectedBatchProcessorTabName==='submitBatch'">
-            <p>This is the submit batch processor tab</p>
+            <div id="selectRegionID">
+                <label>Select State / Region:</label>
+
+                <select ng-model="selectedRegion" ng-options="x.name for x in vm.regionList | orderBy: 'name'"></select>
```
amedenblik:
```suggestion
                <select ng-model="selectedRegion" ng-options="region.name for region in vm.regionList | orderBy: 'name'"></select>

```

Improves readability

src/Views/batchprocessor.html
```diff
@@ -15,7 +15,24 @@ <h4 class="modal-title">StreamStats Batch Processor</h4>
...
         <div id="submitBatch" ng-if="vm.selectedBatchProcessorTabName==='submitBatch'">
-            <p>This is the submit batch processor tab</p>
+            <div id="selectRegionID">
+                <label>Select State / Region:</label>
+
+                <select ng-model="selectedRegion" ng-options="x.name for x in vm.regionList | orderBy: 'name'"></select>
+            </div>
+            <div>                
+                <span>                
+                    <input id="cbDelineate" , type="checkbox" ng-checked="selectedRegion" onclick="return false;"
```
amedenblik: When a state/region has been selected, we want the checkbox to be disabled. Rather than `onclick="return false;"`, let's use the `ng-disabled` directive: https://www.w3schools.com/angular/ng_ng-disabled.asp

src/Views/batchprocessor.html
amedenblik: Include a "value" for the `select` element. The values should be the region codes (ex. "MN" for Minnesota). This is the value we'll want to send to the batch processor services.

amedenblik (changes requested):

harper-wavra (approved):

src/Views/batchprocessor.html
```diff
@@ -15,7 +15,25 @@ <h4 class="modal-title">StreamStats Batch Processor</h4>
     <div style="padding-top: 10px;">
 
         <div id="submitBatch" ng-if="vm.selectedBatchProcessorTabName==='submitBatch'">
-            <p>This is the submit batch processor tab</p>
+            <div id="selectRegionID">
+                <label>Select State / Region:</label>
+
+                <select ng-model="selectedRegion" ng-options="region.name for region in vm.regionList | orderBy: 'name'" value="region.code"></select>
```
amedenblik:
```suggestion
                                <select ng-model="selectedRegion" ng-options="region.code as region.name for region in vm.regionList track by region.code"></select>

```

Your change was showing these value="object:XX" values rather than the code: 
image

The suggested change is one we used in the SCRunoff branch: https://github.com/USGS-WiM/StreamStats/blob/SC-Runoff/src/Views/scstormrunoff.html#L107 (the SCRunoff branch has several forms and may be helpful for future inspiration!)

amedenblik (changes requested):

amedenblik (approved):

PR #1554, opened 2023-02-18
Fixes #1458: Add flow statistics checklist

### LINKED ISSUE #1458: BP: Add "Compute Flow Statistics" checklist
Part of #1455 

- [x] Create a checkbox that says "Compute Flow Statistics". When the checkbox is checked, the checklist (described below) should appear. When the checkbox is unchecked, the checklist should disappear.
- [x] Create a checklist that says "Select Flow Statistics:" 
- [x] When a Region/State is selected, make a service call to https://streamstats.usgs.gov/nssservices/Regions/AZ,NA/Scenarios (example is for Arizona; you include the "code" for the Region and then add ",NA" as well to include nation-wide Statistic Groups) to get a full list of all the Statistics Groups
- [x] Display the "statisticGroupName" of each Statistic Group in the checklist
- [x] Include a way (buttons?) to "Check all" and "Uncheck all" the flow statistics in the checklist

laws-spatial:
Closes #1458.

### DIFF
--- a/src/Controllers/BatchProcessorController.ts
+++ b/src/Controllers/BatchProcessorController.ts
@@ -48,6 +48,8 @@ module StreamStats.Controllers {
         public AppVersion: string;  
         public submitBatchInfo: string;
         public regionList: Object;
+        public flowStatsList: Object;
+        public selectedRegion: any;
 
         //Constructor
         //-+-+-+-+-+-+-+-+-+-+-+-
@@ -81,7 +83,19 @@ module StreamStats.Controllers {
             // call an ng.IPromise first
             this.nssService.getRegionList().then(
                 // set regionList to values of promised response
-                response => { this.regionList = response}
+                response => { this.regionList = response }
+            );
+        }
+        
+        
+
+        // send selected region code and retrieve flows stats list
+        public getFlowStats(rcode:string): void {
+
+            // console.log(rcode)
+            this.nssService.getFlowStatsList(rcode).then(
+                // set flowStatsList to values of promised response
+                response => { this.flowStatsList = response}
             );
         }
 
@@ -90,7 +104,7 @@ module StreamStats.Controllers {
         private init(): void {   
             //console.log("in about controller");
             this.AppVersion = configuration.version;
-            this.getRegions()
+            this.getRegions();
         }
         
 
--- a/src/Controllers/BatchProcessorController.ts
+++ b/src/Controllers/BatchProcessorController.ts
@@ -85,7 +85,8 @@ module StreamStats.Controllers {
             // call an ng.IPromise first
             this.nssService.getRegionList().then(
                 // set regionList to values of promised response
-                response => { this.regionList = response }
+                response => { this.regionList = response; }
+
             );
         }
         
--- a/src/Controllers/BatchProcessorController.ts
+++ b/src/Controllers/BatchProcessorController.ts
@@ -99,6 +99,9 @@ module StreamStats.Controllers {
             );
         }
 
+        // possiblemethod for check/uncheck all
+        // https://makitweb.com/check-uncheck-all-multiple-checkboxes-with-angularjs/
+
         // Helper Methods
         // -+-+-+-+-+-+-+-+-+-+-+-
         private init(): void {   
--- a/src/Controllers/BatchProcessorController.ts
+++ b/src/Controllers/BatchProcessorController.ts
@@ -50,6 +50,7 @@ module StreamStats.Controllers {
         public regionList: Object;
         public flowStatsList: Object;
         public selectedRegion: any;
+        public flowStatsAllChecked: boolean;
 
         //Constructor
         //-+-+-+-+-+-+-+-+-+-+-+-
@@ -99,9 +100,10 @@ module StreamStats.Controllers {
             );
         }
 
-        // possiblemethod for check/uncheck all
-        // https://makitweb.com/check-uncheck-all-multiple-checkboxes-with-angularjs/
-
+        // uncheck/check all flow statistics
+        public toggleflowStatsAllChecked(): void {
+            this.flowStatsAllChecked = !this.flowStatsAllChecked
+        }
         // Helper Methods
         // -+-+-+-+-+-+-+-+-+-+-+-
         private init(): void {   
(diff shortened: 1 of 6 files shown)


src/appConfig.js
```diff
@@ -80,7 +80,8 @@ configuration.queryparams =
         'GageStatsServicesBounds': '/stations/Bounds?xmin={0}&xmax={1}&ymin={2}&ymax={3}&geojson=true',
         'FlowAnywhereEstimates': '/models/FLA/estimate?state={0}',
         'FlowAnywhereGages': '/arcgis/rest/services/IowaStreamEst/FlowAnywhere/MapServer/1/query?geometry={0},{1}&geometryType=esriGeometryPoint&inSR=4326&spatialRel=esriSpatialRelIntersects&outFields=regions_local.Region_Agg,reference_gages.site_id,reference_gages.site_name,reference_gages.da_gis_mi2,reference_gages.da_pub_mi2,reference_gages.lat_dd_nad,reference_gages.long_dd_na&returnGeometry=false&returnIdsOnly=false&returnCountOnly=false&returnZ=false&returnM=false&returnDistinctValues=false&f=pjson',
-        'NSSRegions': '/regions/'
+        'NSSRegions': '/regions/',
+        'NSSRegionScenarios': '/regions/{0},NA/scenarios/'
```
amedenblik: I realized that instead of the URL I provided in the issue, we can actually use a URL that already exists in `configuration.queryparams`:  `'statisticsGroupParameterLookup': '/scenarios?regions={0},NA&statisticgroups={1}&regressionregions={2}',`. It should produce the same results.

The region will obviously go in {0}, but you can just ignore optional parameters {1} and {2} by setting them equal to empty strings. 

Then we can omit line 84.

src/appConfig.js
```diff
@@ -80,7 +80,8 @@ configuration.queryparams =
         'GageStatsServicesBounds': '/stations/Bounds?xmin={0}&xmax={1}&ymin={2}&ymax={3}&geojson=true',
         'FlowAnywhereEstimates': '/models/FLA/estimate?state={0}',
         'FlowAnywhereGages': '/arcgis/rest/services/IowaStreamEst/FlowAnywhere/MapServer/1/query?geometry={0},{1}&geometryType=esriGeometryPoint&inSR=4326&spatialRel=esriSpatialRelIntersects&outFields=regions_local.Region_Agg,reference_gages.site_id,reference_gages.site_name,reference_gages.da_gis_mi2,reference_gages.da_pub_mi2,reference_gages.lat_dd_nad,reference_gages.long_dd_na&returnGeometry=false&returnIdsOnly=false&returnCountOnly=false&returnZ=false&returnM=false&returnDistinctValues=false&f=pjson',
-        'NSSRegions': '/regions/'
+        'NSSRegions': '/regions/',
```
amedenblik:
```suggestion
        'regions': '/regions/',
```

Let's rename this to be more consistent with the names of other queryparams.

src/Controllers/BatchProcessorController.ts
```diff
@@ -48,6 +48,9 @@ module StreamStats.Controllers {
         public AppVersion: string;  
         public submitBatchInfo: string;
         public regionList: Object;
+        public flowStatsList: Object;
+        public selectedRegion: any;
```
amedenblik:
```suggestion
        public selectedRegion: str;

```

We want to avoid "any" went possible

src/Controllers/BatchProcessorController.ts
```diff
@@ -81,16 +84,32 @@ module StreamStats.Controllers {
             // call an ng.IPromise first
             this.nssService.getRegionList().then(
                 // set regionList to values of promised response
-                response => { this.regionList = response}
+                response => { this.regionList = response }
```
amedenblik:
```suggestion
                response => { this.regionList = response; }

```

src/Controllers/BatchProcessorController.ts
```diff
@@ -81,16 +84,32 @@ module StreamStats.Controllers {
...
+        
+        
 
+        // send selected region code and retrieve flows stats list
+        public getFlowStats(rcode:string): void {
+
+            // console.log(rcode)
+            this.nssService.getFlowStatsList(rcode).then(
+                // set flowStatsList to values of promised response
+                response => { this.flowStatsList = response}
```
amedenblik:
```suggestion
                response => { this.flowStatsList = response; }

```

amedenblik (changes requested):
This is looking good! In addition to my small comments, please delete or comment out console.logs just as a good practice.

One thing we can figure out either now or later is that we want each flow statistics checkbox to be associated with its statisticGroupID value, because that numeric value is what we'd like to send to the BP service. We may want to have one big ng-model for all the checkboxes so that later we can easily grab all the statisticGroupIDs of all the checked flow statistics.

harper-wavra (commented):
Other than the comments Andrea left, looks and works well. 

If you want an example of an ng-model for the checkboxes, I would check out the sidebarview.html. Lines 200-213 show something similar to what we are trying to do. (and if you are curious where that comes into play after you delineate a basin and click continue its the basin characteristic list that shows up in the [sidebar](https://user-images.githubusercontent.com/59063181/220371607-61fc6c32-1f5c-452e-8def-26168cce82bf.png))

src/appConfig.js
laws-spatial: @amedenblik, from looking at the streamstatsservices query params farther up starting with SS, I believe that NSSRegions is consistent as is. Would you still like this to be changed?

src/appConfig.js
laws-spatial: > This is looking good! In addition to my small comments, please delete or comment out console.logs just as a good practice.
> 
> One thing we can figure out either now or later is that we want each flow statistics checkbox to be associated with its statisticGroupID value, because that numeric value is what we'd like to send to the BP service. We may want to have one big ng-model for all the checkboxes so that later we can easily grab all the statisticGroupIDs of all the checked flow statistics.

Building on this comment and @harper-wavra, I tried coding this on Friday evening and didn't have any luck. I'll try again based on @harper-wavra code suggestion just so the feature is ready for the next issue.

src/appConfig.js
amedenblik: Yeah let's go ahead and change it. There are numerous queryparams farther up that use NSS (examples include "RegressionScenarios", "statisticsGroupLookup", "statisticsGroupParameterLookup", "estimateFlows") but we don't preface them with "NSS".

src/appConfig.js
amedenblik: Sounds good, yeah those statisticGroupIDs will be handy for the next issue.

amedenblik (approved):
Nice work on the button! Looks nice

PR #1581, opened 2023-03-27
Fixes #1468 and Fixes #1468

### LINKED ISSUE #1468: Hide "edit basin" toolbar for global watersheds
When you edit a basin, the basin characteristic and regression equation values do not change. This seems to only happen for global watersheds, if it is local it works fine.

We should disable the edit basin button when there is a global delineation. See https://github.com/USGS-WiM/StreamStatsServices/issues/9 and https://github.com/USGS-WiM/StreamStats/issues/538 for more information.

Maybe once we switch to SSHydro this will be an easier fix, if the global vs local comes back client side.

harper-wavra:
- enabled edit basin button for all local delineations
- added GA tag for querying prosper layers

Once this looks good, I can rebuild and place on dev and test for GIS folks to test as well.

### DIFF
--- a/src/Views/sidebarview.html
+++ b/src/Views/sidebarview.html
@@ -104,7 +104,7 @@
                                         <button type="button" class="btn-black" ng-click="vm.studyAreaService.undoEdit()"><span class="fa fa-undo"></span>&nbsp;&nbsp;Undo Edits</button>
                                     </form>
                                     <form style="padding:0 0 10px 0;">
-                                        <button type="button" class="btn-black" ng-click="vm.submitBasinEdits()"><span class="fa fa-check"></span>&nbsp;&nbsp;Submit edited basin</button>
+                                        <button type="button" class="btn-black" ng-click="vm.submitBasinEdits()"><span class="fa fa-check"></span>&nbsp;&nbsp;Submit Edited basin</button>
                                     </form>
                                 </div>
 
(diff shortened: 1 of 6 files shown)


src/Views/sidebarview.html
```diff
@@ -87,8 +87,8 @@ <h5>Supported search strings:</h5>
                                     <button id="clearDelinationButton" type="button" class="btn-red fullwidth" ng-click="vm.resetWorkSpace()" ng-show="vm.studyAreaService.selectedStudyArea.FeatureCollection" ng-disabled="!vm.studyAreaService.canUpdate"><span><i class="fa fa-trash-o"></i>&nbsp;&nbsp;Clear Basin</span></button>
                                 </form>
                                 <form style="padding: 5px 0;">
-                                    <p class="p-info">Basin editing has been disabled, and this feature may be discontinued. If you use this feature, please click "Help" then "Submit a Support Request" to let us know!</p>
-                                    <button style="font-size:13px; line-height:16px; color:#808080;" type="button" class="btn-blue fullwidth" ng-click="vm.studyAreaService.showEditToolbar = !vm.studyAreaService.showEditToolbar" ng-disabled="true"><span class="fa fa-edit"></span>&nbsp;&nbsp;Edit Basin</button>
+                                    <p class="p-info" ng-show="vm.studyAreaService.global">Basin editing is currently only available for local delineations.</p>
```
amedenblik:
```suggestion
                                    <p class="p-info" ng-show="vm.studyAreaService.global">Basin editing is currently only available for small basins. If you use would like to use this feature for large basins, please click "Help" then "Submit a Support Request" to let us know.</p>

```

amedenblik (changes requested):

amedenblik (approved):

PR #1648, opened 2023-06-30
Slider, flood stages, legend, ticks, data labels. Eb 1555 2 #1609 # 1557 # 1555

### LINKED ISSUE #1555: Stage vs. Discharge Plot: flood stages
Previous issue: #1492

- [x] Copy this file (https://www.weather.gov/source/aprfc/crossWalk.json) and save it in the src/Assets folder in the StreamStats repo. We will need this file to convert the USGS station ID numbers to the NWS station ID numbers.
- [x] Back in the GagePageController, use that file to determine the NWS station ID for the USGS station ID
- [x] Make a GET request to `https://water.weather.gov/ahps2/hydrograph_to_xml.php?output=xml&gage=NWSStationID` (NWSStationID will be replaced with the NWS Station ID you already determined. For example: https://water.weather.gov/ahps2/hydrograph_to_xml.php?output=xml&gage=bonm8 is the link for USGS site ID 12340000).
- [x] That GET request will return XML content. Parse the XML and look under tag `<sigstages>`. Retrieve the values associated with tags `<action>`, `<flood>`, `<moderate>`, `<major>`, and `<record>`. These values will be the y-axis value for the horizontal line associated with each flood stage.
- [x] Create a function called `curveLookup` that will determine the corresponding x-axis value for each flood stage. See the existing curveLookup function (line 2930) in the weather.gov source code for inspiration (only lines 2963 to 2981 are relevant for this issue). This function is called on line 3547.
- [x] Use the x and y values to plot a vertical and horizontal line for each flood stage
- [x] The following colors should be used for the lines:
     - Action: rgba(255,255,0,0.7)
     - Flood: rgba(255,153,0,0.7)
     - Moderate: rgba(255,0,0,0.7)
     - Major: rgba(204,51,255,0.7)
     - Record: rgba(102,178,255,0.7)
- [x] Label the lines as seen in the weather.gov plot.

Tips:
- If you go to https://www.weather.gov/source/aprfc/gageAnalysis.html?site=12340000 and right click "View Source", search for the function `addFloodStageRating`.
- You may want to add these lines as HighCharts "plot lines". Here is some relevant documentation:
     - https://www.highcharts.com/docs/chart-concepts/plot-bands-and-plot-lines 
     - x axis plot lines: https://api.highcharts.com/highcharts/xAxis.plotLines
     - y axis plot lines: https://api.highcharts.com/highstock/yAxis.plotLines
     - Demo of y axis plot lines: https://www.highcharts.com/demo/stock/yaxis-plotlines
- Not all USGS gages will be available in the NWS service, so we'll want to handle those cases 
- The NWS service may return undefined values for some/all of the flood stages, so we'll want to handle those cases (see line 3542 of the weather.gov source code)
- Don't hesitate to reach out with questions!

Next issue: #1556

### LINKED ISSUE #1557: Stage vs. Discharge Plot: USGS measured data sliders
Previous issue: #1556 

- [x] Add a slider to filter USGS measurements values by a range of months
- [x] Add a slider to filter USGS measurements values by range of years

Tips:
- I would recommend trying rzsliders. The [wateruse.html file](https://github.com/USGS-WiM/StreamStats/blob/dev/src/Views/wateruse.html#L20) uses rzslider elements, which are described here: https://angular-slider.github.io/angularjs-slider/. You can see an example of how this looks by zooming to Ohio in the client, selecting "Ohio" in the left sidebar, delineating a basin, and then clicking the "Check for Water Use" button in the left sidebar:
image
- When the sliders are changed, you may want to try using a Javascript filter to filter the USGS measurement data: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

### LINKED ISSUE #1609: Add Legend for discharge plot
create legend for #1556

swimbott:
This should contain the code for a number of issues listed in the title. 

There is still an issue with the dynamic ticks when Log Scale is toggled. The legend currently hides the flood stage labels, but once the log scale tick issue is resolved, the legend I think will be appearing in an appropriate place. 

Lines 3083-3098 has the tick positioner code along with 3020-3026.


src/appConfig.js
```diff
@@ -5,11 +5,11 @@ configuration.environment = 'development';
 configuration.baseurls =
     {
         'NWISurl': 'https://waterservices.usgs.gov/nwis',
-        'StreamStatsServices': 'https://test.streamstats.usgs.gov',
+        'StreamStatsServices': 'https://streamstats.usgs.gov',
```
harper-wavra: I don't think there should be any changes to this file.

src/Controllers/GagePageController.ts
```diff
@@ -248,8 +248,20 @@ module StreamStats.Controllers {
         public formattedDailyPlusAvg = [];
         public formattedDischargePeakDates = []; // Stage vs. Discharge Plot
         public dailyValuesOnly = [];
-        public ageQualityData = 'age'; //Stage vs. Discharge Plot
-        public error: any; //Stage vs. Discharge Plot
+        public ageQualityData = 'age';
+        public error: any;
```
harper-wavra, line 252: I think a more descriptive label for this variable might be helpful.

src/Controllers/GagePageController.ts
```diff
@@ -248,8 +248,20 @@ module StreamStats.Controllers {
...
-        public error: any; //Stage vs. Discharge Plot
+        public ageQualityData = 'age';
+        public error: any;     
+        public monthSliderOptions: any;
+        public startMonth: number;
+        public endMonth: number;
+        public startYear: number;
+        public endYear: number;
+        public yearSliderOptions: any;
+        public showQuality: any;
```
harper-wavra: I don't think this variable is used.

src/css/main.css
```diff
@@ -7,1987 +7,2034 @@
...
+             display: flex;
+             align-items: center;
+           }
+           
+           .ageQualityLegend-color {
+             display: inline-block;
+             width: 12px;
+             height: 12px;
+             margin-right: 5px;
+           }
```
harper-wavra: I don't think this is used.

src/css/main.css
```diff
@@ -7,1987 +7,2034 @@
...
+     display: inline !important;
+ }
+ 
+ .disclaimerReport p {
+     max-width: 100%;
+     margin-bottom: auto;
+     padding-top: 5px;
+ }
+ 
+ #printArea span:not(rzslider span) {
```
harper-wavra: We use `printArea` in other places so we just want to avoid any unknown consequences from editing this. I think I checked most of them & it was fine, I guess I am just wondering what this change was for?

I think it used to be: 
```
#printArea span {
padding: 0 0 0 3px !important;
}
```


</body>
</html>

src/Controllers/GagePageController.ts
```diff
@@ -2871,309 +2947,456 @@ module StreamStats.Controllers {
...
+                symbol: 'diamond',
+                radius: 3
+            },
+            showInLegend: this.error == false //!this.measuredObj.every(item => isNaN(item.y)) 
+        }
+    ]},
+    
+    // Format flood stages as data series
+    this.formattedStages = []
+    this.stages.forEach((stage, index) => {
```
harper-wavra, line 3182: Sometimes there are no stages which causes this to print out a console error, see gage 06359800 as an example. If we wrap this in an if statement that should stop those console errors.

```
if (this.stages) {
    lines 3193-3253
}
```

harper-wavra (changes requested):
This plot is really coming together 🎊 !

src/Controllers/GagePageController.ts
swimbott: Agreed! However, changing error to a different name and it's associated references, was screwing things up. Not sure if I should spend more time on this and figure it out?

src/appConfig.js
swimbott: Not sure why it was showing test for you as I didn't change it. Maybe during the transfer with WSL, this changed that?

src/Controllers/GagePageController.ts
swimbott: From what I can tell, there are multiple references to these variables?

harper-wavra: Sorry not sure why it highlighted them all, I just mean showQuality doesn't seem used. 
image

src/Controllers/GagePageController.ts
harper-wavra: mm not sure why that would happen. Did you try changing:

line 252 to `public measuredError: any;`
line 1090 to `this.measuredError = data.includes(errorMessage)`
line 1091 to `if (this.measuredError == false) {`
line 3189 to `showInLegend: this.measuredError == false`

Only those lines should need to be changed.

src/appConfig.js
harper-wavra: I guess I am not sure how they got changed but we need to change them back to https://github.com/USGS-WiM/StreamStats/blob/gageplots/src/appConfig.js#L5-#L19. These three lines should be test.streamstats.

amedenblik (changes requested):
This is looking good! Here are some things I noticed:

- [x] The vertical flood stage labels were cutting off on my screen:
image

- [x] I am not able to zoom in such that the y-axis is changed. If I draw a tiny box to zoom in, the y-axis minimum and maximum value stay the same.

- [x] The legend can cover up the data at times (this is an example for gage 12340000):
image

- [x] I am not sure all the expected flood stages are showing up. Compare this graph of 12340000 to the weather.gov graph (note the lines at the top of the graph):
image
image

src/Controllers/GagePageController.ts
swimbott: Yea those exact lines after finding all references. I change it to any other variable name and then I lose my age and quality buttons.

amedenblik (changes requested):
Awesome work addressing those comments! The changes look great. 

Unless @harper-wavra has further comments, the only thing left is to resolve merge conflicts with the GagePageController.ts and .js:

image

While in branch EB-1555-2, enter `git pull origin gageplots`. Then use VS code to resolve the merge conflicts with the GagePageController.ts file. Let us know if you need any help with that! After resolving the merge conflicts, run `npm start` again to carry over changes to the associated .js file, and then check to make sure everything still works as you expect.

harper-wavra (approved):

amedenblik (approved):
Looks good!

amedenblik:

PR #1649, opened 2023-06-30
Fixes #1626

### LINKED ISSUE #1626: File upload button should be un-highlighted after a user uploads a file

harper-wavra:
I think we just needed to add the other type of zip (application/x-zip-compressed) to the html if statements. 

Let me know if this doesn't work for anyone.


alaws-USGS (approved):
The attach shapefile box is no longer yellow after upload. Great job.

amedenblik (approved):
Works for me too!

amedenblik:
Updated https://dev.streamstats.usgs.gov/bp/

PR #1683, opened 2023-07-13
Fixes #1462 - add url params for BP modal

### LINKED ISSUE #1462: BP: Create URL parameter
Part of #1455 

- [x] Add a URL path such as https://streamstats.usgs.gov/ss/bp/submit, so that when a user accesses that link, it automatically opens the BP modal and goes to the "Submit Batch" tab.
- [x] Add a URL path such as https://streamstats.usgs.gov/ss/bp/status, so that when a user accesses that link, it automatically opens the BP modal and goes to the "Batch Status" tab.
     - [x] Add a URL path parameter https://streamstats.usgs.gov/ss/bp/status?email=person@example.com, so that when a user accesses that link, it automatically opens the BP modal, goes to the "Batch Status" tab, and filters the results by the requested email address.
- [x] Add a URL path such as https://streamstats.usgs.gov/ss/bp/streamgrid, so that when a user accesses that link, it automatically opens the BP modal and goes to the "Download Stream Grids" tab.

harper-wavra:
These are the url params I added. They should show up as you switch between tabs and be removed if you close the modal.
- /?BP=submitBatch
- /?BP=batchStatus
- /?BP=batchStatus?email=person@example.com
- /?BP=streamGrid
- /?BP=manageQueue


amedenblik (approved):
Nice work!

Issue #1686, opened 2023-07-14
Update BP Service URLs ( after deploy )


harper-wavra:
When we have deployed the BP services to a dev and production environment we need to update the BP URLs.

-  [x] Update in the appconfig. 
    - [x] Dev & Test- [Line 21](https://github.com/USGS-WiM/StreamStats/blob/batch-processor/src/appConfig.js#L21) should point to the devURL. We are disabling the BP on test #1763 so line 21 can always point to the URL.
    - [x] Production - we will need to add a new line `if (window.location.host === 'streamstats.usgs.gov')` to point to the production URL.
    - [x] Update Stream Grids https://github.com/USGS-WiM/StreamStats/issues/1768
- [ ] Update the manage queue URLs. 
    - [x] Dev- [update line 901](https://github.com/USGS-WiM/StreamStats/pull/1687/files#diff-f4968787b9d4ca9c35f7a9cec7cd0785448148fcfbe838b2e2e658a96461b006R901)
    - [x] Production - [update line 897](https://github.com/USGS-WiM/StreamStats/pull/1687/files#diff-f4968787b9d4ca9c35f7a9cec7cd0785448148fcfbe838b2e2e658a96461b006R897)
    - [ ] Check to make sure the pause, deletes, order queue, are hitting the correct service urls for the queue selected. It should work but I wasn't able to 100% check with #1677

harper-wavra:
Related issue #1763

PR #1687, opened 2023-07-14
Fixes #1677 

### LINKED ISSUE #1677: Select server for Batch Processor Manage Queue tab
Create a way for the user to specify which Batch Processor queue they would like to manage:

At least 2 options for this:

1. On the Manage Queue tab, add a drop down with options for Development Server or Production Server. When the user selects the server, the service URLs will need to be changed
2. Use separate URLs for each server: examples
      - https://apps-int.usgs.gov/streamstats/dev/bp/
      - https://apps-int.usgs.gov/streamstats/bp/

harper-wavra:
Added dropdown in manage queue to select which queue you want to manage.

Since we don't have production instance in the DB yet the production queue returns nothing. I added comments about where to switch the urls when we have the environments all set up & created an issue to update the URLS https://github.com/USGS-WiM/StreamStats/issues/1686.

### DIFF
--- a/src/Controllers/BatchProcessorController.js
+++ b/src/Controllers/BatchProcessorController.js
@@ -80,6 +80,8 @@ var StreamStats;
                 _this.retrievingManageQueue = false;
                 _this.flowStatIDs = [];
                 _this.submitBatchOver250 = false;
+                _this.queues = ['Production Queue', 'Development & Test Queue'];
+                _this.selectedQueue = "Production Queue";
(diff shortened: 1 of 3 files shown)


amedenblik (approved):

amedenblik:
Deployed to:
- https://dev.streamstats.usgs.gov/bp/
- https://staging-apps.usgs.gov/streamstats/bp/

PR #1692, opened 2023-07-14
Closes #1688, #1689, #1691 Batch Processor improvements

### LINKED ISSUE #1688: Add refresh button to Batch Status and Manage Queue tabs
Add a refresh button to the Batch Status and Manage Queue tabs so that the user can easily see the latest queue/results

### LINKED ISSUE #1689: Use the Enter button on Batch Status tab
When the user enters an email address and then presses the enter button, it should work the same as clicking the "Check Status" button

### LINKED ISSUE #1691: Add asterisk for US National region Basin Characteristics
In the list of Basin Characteristics, add an asterisk to the note about BCs not always being available.

Add an asterisk after each BC label for which this applies: all BCs that are associated with the US National region that are _not_ also associated with the selected State/Region.

amedenblik:
Closes #1688, #1689, #1691

### DIFF
--- a/src/Controllers/BatchProcessorController.js
+++ b/src/Controllers/BatchProcessorController.js
@@ -82,6 +82,7 @@ var StreamStats;
                 _this.submitBatchOver250 = false;
                 _this.queues = ['Production Queue', 'Development & Test Queue'];
                 _this.selectedQueue = "Production Queue";
+                _this.isRefreshing = false;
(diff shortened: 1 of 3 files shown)


harper-wavra (approved):
Had to change StreamStatsServices in the appconfig to point to prod since the test StreamStatsServices are down right now, but they look good.

PR #1714, opened 2023-08-18
Fixes #1703 - Reorder Queue Button

### LINKED ISSUE #1703: Disable Reorder Queue button when there are no batches in the queue
image

The Reorder Queue button should be disabled because there are no batches that are able to be reordered

Low priority issue

harper-wavra:
Only enable reorder queue button if more than 1 batch has status of 'In Queue'.


amedenblik:
Check the target branch :) I would change it, but it gave me some warnings

harper-wavra:
Fixed 👍

amedenblik (approved):

PR #1717, opened 2023-08-18
Fixes #1670 - add sediment region layer

### LINKED ISSUE #1670: Display Sediment Regions map layer
Display the Sediment Regions map layer in Minnesota. The SedimentRegions_shapefile is included in the ScienceBase data release: https://www.sciencebase.gov/catalog/item/61572adcd34e0df5fb9f8300. Will is working on updating the ScienceBase item to serve this is an map service. Once that is complete (probably will take several weeks), we can work on this issue.

harper-wavra:
- added MN sediment region layer

I made both layers `"visible": false` on initial click of Minnesota. The river size class layer wasn't too annoying but the sediment region layer kind of is & I was thinking most users probably won't care about these layers. Maybe we could have a note in the sediment modal telling users they can turn on these layers if they would be helpful?

### DIFF
--- a/src/appConfig.js
+++ b/src/appConfig.js
@@ -425,6 +425,17 @@ configuration.regions = [
                             case 'SE': return {color: "rgb(255, 255, 190)", fillOpacity: .80};
                             case 'SW': return {color: "rgb(215, 215, 158)", fillOpacity: .80};
                         }
+                    },
+                    onEachFeature: function (feature, layer) {
+                        var popupContent = '<h5>MN Sediment Regions</h5> ';
+                        var queryProperties = { 
+                            "US_L4NAME": "Sediment Region", 
+                            "Region": "Region"
+                        };
+                        Object.keys(queryProperties).map(function (k) {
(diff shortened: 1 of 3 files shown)


src/appConfig.js
```diff
@@ -410,7 +410,47 @@ configuration.regions = [
                         "label": "Great River (GR)"
                     }]
                 }],
-
+            },
+            "SedimentRegions": {
```
amedenblik, line 414: image
 I would recommend changing this popup to:

**MN Sediment Regions**
Sediment Region: Southwest
Region Abbreviation: SW

amedenblik (changes requested):
Yeah, I think that's a good idea to keep the layers off by default. Maybe later we can turn the river size class layer on if the user enters the SedStats workflow.

src/appConfig.js
harper-wavra: Thanks for catching that! I was thinking if I didn't have the query properties there wouldn't be a popup at all.

amedenblik (approved):

amedenblik:
Deployed to https://dev.streamstats.usgs.gov/mn-sediment/

PR #1729, opened 2023-08-23
Fixes #1727

### LINKED ISSUE #1727: Prevent deletion of batches when In Progress
Hide the trash can when a batch is In Progress

harper-wavra:
Don't allow deletion when batch status is in progress.


amedenblik (changes requested):
We need to also prevent batch deletion on the "Batch Status" tab: 

image

amedenblik (approved):

amedenblik:
Deleting batches on the Batch Status tab isn't working. We'll address in a new issue.

image
image

PR #1741, opened 2023-08-29
Fixes #1740

### LINKED ISSUE #1740: BP - Check if scenarios are available
We need to check if `ScenariosAvailable=false` in the appconfig. If it equals false, there should be no options for flow statistics. 

For example, if I select Connecticut River Basin in the BP I am currently able to select from 3 flow statistics. None should be available for selection because in the appconfig under CRB `"ScenariosAvailable": false`.

harper-wavra:
disable flow statistic selection if scenarios are not available


src/Views/batchprocessor.html
```diff
@@ -143,6 +143,10 @@ <h4 class="modal-title">StreamStats Batch Processor</h4>
             <div ng-if="vm.cbFlowStats" style="padding-top: 15px;" class="well well-sm">
                 <h3 class="collapsible" ng-click="vm.collapseSection($event, 'flowStatistics')"><i ng-class="!vm.flowStatsCollapsed ? 'fa fa-chevron-right': 'fa fa-chevron-down'"></i>&nbsp;&nbsp;Select Flow Statistics</h3>
                 <div class="collapsible-content" style="background-color: #f5f5f5;">
+                    <!-- Message that no flow statistics are available -->
+                    <div ng-if="vm.flowStatsList.length == 0">
+                        <span>Scenario computation has not been enabled yet for selected region.</span>
```
amedenblik, line 148: Maybe we get rid of the "yet", which implies that scenarios may be available later? 

```suggestion
                        <span>Scenario computation has not been enabled for selected region.</span>

```

amedenblik (changes requested):
Works well!

src/Views/batchprocessor.html
harper-wavra: I copied the text we are using in the sidebar when scenarios are not available, should we switch that too.

amedenblik: Ah didn't realize, let's just stick with this to be consistent!

amedenblik (approved):

Issue #1755, opened 2023-09-06
Disable BP tabs besides Manage Queue if on NatWeb


amedenblik:
When on https://staging-apps.usgs.gov/streamstats/bp or https://apps-int.usgs.gov/streamstats/bp, disable all Batch Processor tabs except the Manage Queue tab. This way, this site will only be used for managing the queue and we won't accidentally submit batches to the wrong queue (dev/test/prod). 

Use an appconfig boolean for manageQueue

PR #1759, opened 2023-09-06
Fix manage queue bug related to #1755

amedenblik:
I missed a bug in #1756 where if manageBPqueue=true and no URL parameters were sent, the Submit Batch tab would open when opening the Batch Processor modal.


harper-wavra (approved):

amedenblik:
Updated https://staging-apps.usgs.gov/streamstats/bp

PR #1770, opened 2023-09-14
Fixes #1761 Update MA Culverts map services

### LINKED ISSUE #1761: Update MA Culverts map service
Switch the map service to the new ScienceBase map service that Luke has created: https://services.arcgis.com/v01gqwM5QqNysAAi/arcgis/rest/services/Massachusetts_Stream_Crossing_Spatial_Data/FeatureServer

>  I've spun up a WMS with the formatted lookup table to store the data similar to how StreamStats stores basin characteristic tables. The point locations still have all the attributes in the table so you should be able to use those the same way you currently are but let me know if you are able to use the lookup table or if it is even worth switching to use a lookup table like that.

amedenblik:
Closes #1761

I commented out the `getCulvertCreds()` function instead of deleting in case the ScienceBase map service goes away and we have to go back to using it.


harper-wavra:
I don't entirely remember why he switched to using a lookup table, 'formatted lookup table to store the data similar to how StreamStats stores basin characteristic'. But did we need to look into this?

harper-wavra (approved):

amedenblik:
I missed that- yes, we should look into that! I see the table in the map service here: https://services.arcgis.com/v01gqwM5QqNysAAi/ArcGIS/rest/services/Massachusetts_Stream_Crossing_Spatial_Data/FeatureServer/2

But I am not sure how/why to use it. That table does not have all the same attributes as the Stream Crossings layer. I suggest we merge this pull request and I'll open a new issue for the table (#1771) and email him.

amedenblik:
Deployed to https://dev.streamstats.usgs.gov/ma-culverts/ and emailed Luke with an update.

PR #1772, opened 2023-09-14
Fixes #1763

### LINKED ISSUE #1763: Disable BP on TestWeb
We are not going to allow batch processing on test. If the host url is test.streamstats we should disable the BP button.

harper-wavra:
add disable bp button option in appconfig


src/appConfig.js
```diff
@@ -6,6 +6,7 @@ configuration.warningModalMessage = "Due to heavy demand, StreamStats is current
 configuration.showBPWarning = false;
 configuration.warningBPMessage = "Due to heavy demand, StreamStats is currently experiencing system interruptions. If you receive errors, please try back again later.<br><br>Thank you for your patience."
 configuration.manageBPQueue = false;
+configuration.showBPButton = true;
 
 configuration.baseurls =
```
amedenblik, line 15: What if we added this? Not sure if it would work 

```
if (window.location.host === 'test.streamstats.usgs.gov') {
     configuration.showBPButton = false;
}
```

amedenblik (approved):
Looks good to me. Update changelog and rebuild and should be good to go

harper-wavra:
Updated dev, test and the internal manage queue site.

Issue #1707, opened 2023-07-25
Gageplots Feedback - Annual Peak Streamflow plot


amedenblik:
More feedback may be added later! This is what we have received so far.

- [x] Consider adding option for a log scale
- [x] It should be noted that the % AEP flows are the at-site stats as opposed to the weighted %AEP flows (weighting of at-site values with regression values)
     - Maybe add a dropdown for data sources for each plot -- this way users can see exactly what stats/data is being used (see #1709)
     - (Awaiting guidance from Pete M on the wording of the note)
- [x] "It’s confusing that the x-scale shown for the peak flows is only correct for the default window"
     - [x] We're going to try: if the x-scale is zoomed to be greater than 1 year, plot the points on their actual year & uncheck the "Plot Annual Streamflow Peaks on Latest Year" box  
     - [x] Consider allowing an option to plot the peak flows on any user-specified year: drop-down options would be to plot points on actual date, 2023, 2022, 2021, etc... 
- When you use the slider at the bottom of the graph, all the peak flows get pushed to the far right. It would be better if only the peak flows associate with the period chosen on the slider would show up and use the x-scale being used for the daily mean streamflow.
     - See previous point 
- [x] For some sites (06334330 for example) the annual peak streamflow plot loads in blank and only by zooming out the slider can one see the peak flow values. Looks like the peaks are plotting on the wrong year for some reason:
![image](https://github.com/USGS-WiM/StreamStats/assets/40237491/a67fc1ab-d934-4bb6-b0fd-9c5c8698a79f)
     - This may occur because the most recent year is determined based on the daily streamflow data, which is unavailable for this site

- [x] This one (06127900) seems to be having problems too

- [x] Add these statistics to the drop-down statistic group options:
     - Weighted AEP Statistics: 
          - Weighted 80% AEP: 1065
          - Weighted 66.7% AEP: 20
          - Weighted 50% AEP: 513
          - Weighted 42.9% AEP: 512
          - Weighted 20% AEP: 21
          - Weighted 10% AEP: 22
          - Weighted 4% AEP: 23
          - Weighted 2% AEP: 24
          - Weighted 1% AEP: 25
          - Weighted 0.5% AEP: 26
          - Weighted 0.2% AEP: 27
     - Regulated AEP Statistics:
          - Regulated 80% AEP: 1328
          - Regulated 66.7% AEP: 517
          - Regulated 50% AEP: 516
          - Regulated 42.9% AEP: 518
[...]

PR #1783, opened 2023-10-03
Adds log scale functionality for #1707: Gageplots Feedback - Annual Peak Streamflow plot

maggiejaenicke:
Seems like the `containsNegatives()` function was deleted at some point. Adding that back to allow log scale checkbox to operate again per request in #1707. 

Does not close issue -- leave #1707 open.


harper-wavra (approved):
Oh shoot, I wonder when that happened. Looks good though!

Issue #1709, opened 2023-07-25
Gageplots Feedback - General


amedenblik:
More feedback may be added later! This is what we have received so far.

- [x] Add gage number to title of plots ([notes here](https://wim-usgs.slack.com/archives/C2PG1BC3S/p1690570119823459)) #1787
 - [ ] Disable hamburger button options except for View in Full Screen
      - See [these Highcharts settings](https://api.highcharts.com/highcharts/exporting.buttons.contextButton.menuItems)
 - [ ] Fix small issues in appearance when you click the "Print" button at bottom of modal
      - All flood stats lines are showing up -- we want it to just be the selected lines
 - [x] Add a collapsible section for each plot that lists the data sources #1787
      - [x] Say "Not available" if a data sources is not available for that site (example: 06177020 doesn't have daily streamflow)

maggiejaenicke:
Add to all three plots: 

In chartConfig constructor: 
`exporting : { buttons: {contextButton: {menuItems: string[]}}}
`

In this.chartConfig in `createAnnualFlowPlot()` function:
```
               exporting: {
                    buttons: {
                        contextButton: {
                            menuItems: ['viewFullscreen']
                        }
                    }
                }
```

Other note: the export options do work well for the Annual Peak Streamflow plot so we could always evaluate which options to keep on a plot-by-plot basis.

amedenblik:
Closed in favor of #1970 and #1971

PR #1787, opened 2023-10-05
Addresses 2 tasks in #1709: Gageplots Feedback - General

maggiejaenicke:
- Adds gage numbers to each plot title.
- Adds data sources to each plot, including "Not Available" message when no data at site for each source.

Does not close full issue -- keep #1709 open.

### DIFF
--- a/src/Views/gagepage.html
+++ b/src/Views/gagepage.html
@@ -241,7 +241,14 @@
                                 <input type="checkbox" ng-if="vm.formattedFloodFreq.length > 0 || vm.formattedAltFloodFreq.length > 0" id="toggleFloodStats" name="toggleFloodStats" ng-model="vm.showFloodStats" ng-change="vm.toggleFloodStats()" checked>
                                 <label for="toggleFloodStats" ng-if="vm.formattedFloodFreq.length > 0 || vm.formattedAltFloodFreq.length > 0" > Show Flood Statistics:</label>
                                 <select id="gageFloodStatsSelect" ng-if="vm.formattedFloodFreq.length > 0 || vm.formattedAltFloodFreq.length > 0 " style="width:31%; text-align-last:center; margin-top: 5px;" name="gageFloodStatsSelect" ng-model="vm.selectedFloodFreqStats" ng-options="floodFreqStat.name for floodFreqStat in vm.allFloodFreqStats" ng-change="vm.chooseFloodStats()" ng-disabled="vm.showFloodStats == false"></select><br> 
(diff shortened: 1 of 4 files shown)


src/Views/gagepage.html
```diff
@@ -240,16 +240,35 @@ <h4 class="modal-title">StreamStats Gage Page</h4>
...
-                            </div>
-                            </div>
+                                <select id="gageFloodStatsSelect" ng-if="vm.formattedFloodFreq.length > 0 || vm.formattedAltFloodFreq.length > 0 " style="width:31%; text-align-last:center; margin-top: 5px;" name="gageFloodStatsSelect" ng-model="vm.selectedFloodFreqStats" ng-options="floodFreqStat.name for floodFreqStat in vm.allFloodFreqStats" ng-change="vm.chooseFloodStats()" ng-disabled="vm.showFloodStats == false"></select>
+                                <div class="data-sources">
+                                    Data: 
+                                    <a ng-if="vm.formattedPeakDates.length > 0" href="https://nwis.waterdata.usgs.gov/usa/nwis/peak/?format=rdb&site_no={{vm.gage.code}}">NWIS Annual Peak Streamflow,</a>
+                                    <text ng-if="vm.formattedPeakDates.length === 0">NWIS Annual Peak Streamflow (Not Available),</text>
+                                    <a ng-if="vm.formattedDailyFlow.length > 0" href="https://nwis.waterservices.usgs.gov/nwis/dv/?format=json&sites={{vm.gage.code}}&parameterCd=00060&statCd=00003&startDT=1900-01-01">NWIS Mean Daily Streamflow,</a>
```
amedenblik: "Mean Daily Streamflow" doesn't match what's in the legend, which is "Daily Mean Streamflow". I actually prefer "Mean Daily Streamflow" though-- can you please change the legend text?

src/Views/gagepage.html
```diff
@@ -240,16 +240,35 @@ <h4 class="modal-title">StreamStats Gage Page</h4>
...
+                                    <text ng-if="vm.formattedDailyFlow.length === 0">NWIS Mean Daily Streamflow (Not Available),</text>
+                                    <a ng-if="vm.formattedFloodFreq.length > 0 || vm.formattedAltFloodFreq.length > 0 " href="https://test.streamstats.usgs.gov/gagestatsservices/stations/{{vm.gage.code}}">USGS Flood Frequency Statistics,</a>
```
amedenblik: We shouldn't hardcode `test.streamstats.usgs.gov` because when we move to Production, we'll want to use `streamstats.usgs.gov`. You'll want to use the value of `configuration.baseurls.GageStatsServices`, since that will automatically use the correct base URL.

Thinking bigger picture-- in case the URLs ever change and just to reduce duplicity or potential errors, a better practice for this entire section might be to set the URLs as variables in the GagePageController and then refer to them in the gagepage.html file.

amedenblik (changes requested):

amedenblik (approved):
Looks good!

amedenblik:

PR #1825, opened 2023-11-10
Fixes #1818 Display Batch Processor times in local time

### LINKED ISSUE #1818: Consider time zone other than UTC
Consider using ET, CT, etc. instead of UTC for the time zones

amedenblik:
Closes #1818 

Dates in the Batch Processor Batch Status tab are now displayed in local time like this: 11/9/23 1:42 PM GMT -0400.

There is not a consistent way to provide a time zone code like "ET". Also the times vary due to daylight savings (some of my batches were showing GMT -0400 and some were showing GMT -0500 depending on if thy were submitted before or after the time change). So, I went with just having "GMT -0400" etc. on each line.


harper-wavra (approved):
Working well on the batch status tab. Good to merge unless you think we should we do a similar change for the manage queue tab?

amedenblik:
- Made this change on the manage queue table
- Fixed issue where null datetimes (ex. there was no TimeC ompleted because the batch hadn't finished yet) were showing up as "null GMT null". Now they are just blank as expected.

harper-wavra (approved):
Looks good 👍  

I rebuilt and will push it to the manage queue site and dev

PR #1836, opened 2023-11-18
Fixes #1742 Incorporate excludePolygon endpoint

### LINKED ISSUE #1742: Incorporate excludePolygon endpoint
_Waiting on services_

Since we are using sspourpoint & its exclude polygon endpoint in the BP we should also start using it in the client. I think we can work off the bp branch and push it out with that update.

amedenblik:
Closes #1742 

Exclusion polygons are now evaluated via the https://streamstats.usgs.gov/pourpoint/ssExcludePoly instead of querying the state map services directly


src/Controllers/MapController.ts
```diff
@@ -902,75 +905,45 @@ module StreamStats.Controllers {
...
+                                    this.toaster.pop("warning", "Warning", result.message.text, true, 0);
                                     gtag('event', 'ValidatePoint',{ 'Label': 'Not advised' });
+                                    this.startDelineate(latlng, true, result.message.text);
                                 }
+                                // TODO in future issue: add more types of exlusion polygons
                             }
-
-                            this.cursorStyle = 'pointer';
+                            
                         });
```
harper-wavra, line 966:
```suggestion
                        }, (error) => {
                            //toast message when error              
                        }).finally(() => {  
                            // probably don't need anything in here      
                        });

```

harper-wavra (changes requested):
Works good! 

I think we should just catch the error if there is one. Not sure if we should say there is an error and keep moving forward with the delineate or just stop the whole process right here.

src/Controllers/MapController.ts
amedenblik: If the exclude poly service fails in the BP, we return the point as if it is not in an exclude polygon and proceed with delineation, so I'll plan to replicate that functionality

amedenblik: The `}, (error) => {` line wasn't triggering with my testing, so I also checked the status of the response to catch when the query exclude poly service was failing.

I tested that by changing excludePolygonQuery to this (erroneous POST body):

```
var excludePolygonQuery = {
                "regionXYZ": region,
                "points": [
                    {
                    "lat": lat,
                    "long": lng
                    }
                ]
            }
```

harper-wavra (commented):
Working good. 

I did see this error in the code, but it doesn't seem to be affecting anything. 
image

amedenblik:
> Working good.
> 
> I did see this error in the code, but it doesn't seem to be affecting anything. image

I added a brace to fix that error 👍 It didn't change the JavaScript though and no changes were made after rebuild

harper-wavra (approved):

PR #1840, opened 2023-11-21
Fixes #1833 Add "refresh batch" button to BP Manage Queue

### LINKED ISSUE #1833: Add "refresh batch" button to BP Manage Queue
Add the ability to refresh a batch to the Manage Queue tab. This will help if the services go down and we need to restart a batch.

amedenblik:
In the Manage Queue tab of the Batch Processor:

- You can now click an icon to "refresh" a batch. This will delete the current data associated with the batch and put the batch in the queue.
- There is now a "Start Next Batch" button that will kick off a worker.

Together these tools will allow us to easily restart batches if the server goes down while batches are running.

Set `configuration.manageBPQueue = true;` for testing.


dist/appConfig.js
```diff
@@ -73,6 +73,8 @@ configuration.queryparams =
         'SSBatchProcessorStatusMessages': '/status/',
         'SSBatchProcessorStreamGrids': '/streamgrids/',
         'SSBatchProcessorSubmitBatch': '/batch',
+        'SSBatchProcessorRefreshBatch': '/batch/{0}/refresh?deleteCurrentData={1}',
+        'SSBatchProcessorStartWorker': '/worker',
```
harper-wavra, line 77: Do we need to add these to the other appconfigs?

harper-wavra (approved):
Worked great! Tested on a batch on dev and prod.

dist/appConfig.js
amedenblik: Yes, woops. Totally forgot about that!

PR #1841, opened 2023-11-21
Fixes #1839 Add warning message if BP services are down 

### LINKED ISSUE #1839: Add warning message if BP services are down
If https://streamstats.usgs.gov/batchprocessor/regions/ isn't returning, automatically add a warning message header saying that the BP is currently down.

amedenblik:
Closes #1839 

Quick fix to show a warning message at the top of the BP modal, if the Region endpoint isn't working (which typically indicates that the BP is down).

This isn't really comprehensive (for example, if the user initially goes straight to the Batch Status tab, they will not see the error box). But hopefully it will help explain things a little bit if users go to the Submit Batch tab and see they aren't able to submit a batch because the services are down.

Also:
- Added some padding to the yellow wim-warning box just so it looks nicer
- The appConfig boolean showBPWarning wasn't actually working-- fixed this


harper-wavra (approved):
Looks good 👍 
Tested by changing the region endpoint to a fake one.

PR #1845, opened 2023-11-27
#1844 - fixed repeated regions in streamgrid tab

### LINKED ISSUE #1844: BP Stream Grids tab sometimes shows repeated regions
I think this happens when you visit https://dev.streamstats.usgs.gov/ss?bp=streamGrid directly. I think the regions list is getting populated twice

harper-wavra:
`selectBatchProcessorTab` was getting called twice when you loaded the tab from the URL parameters, which was causing the `loadStreamgrid` function to be called twice. 
`getRegions` would finish after we needed the regions, so I added a one second timeout to wait for that request to finish.


amedenblik:
Resolved merge conflicts and rebuilt

amedenblik (approved):

PR #1847, opened 2023-11-29
#1169 and #1828 Line delineation workflow with basin characteristics

### LINKED ISSUE #1169: Allow multipoint delineation
Waiting on open source delineation. Deadline: end of September 2022

From the [SC StreamStats Phase II proposal](https://doimspp.sharepoint.com/sites/SSPhaseII/Shared%20Documents/Forms/AllItems.aspx?id=%2Fsites%2FSSPhaseII%2FShared%20Documents%2FGeneral%2FProposal%2FSC%5FStreamStats%5FPhase%5FII%5FProposal%5FFinal%2Epdf&parent=%2Fsites%2FSSPhaseII%2FShared%20Documents%2FGeneral%2FProposal):

"Due to data improvements from lidar data in South Carolina, there can be multiple flow lines throughout swampy low-lying areas of the landscape. In these low-lying areas, streamgage operations typically require multiple measurements across various channels to capture the full extent of the drainage basin. The StreamStats application currently does not provide a way to spatially select and aggregate multiple stream segments to capture the full extent of flow in these areas, and a delineation from one ungaged point may not represent the full drainage basin represented by the gage. Multipoint delineation would add the functionality to select multiple stream segments by point selection and provide a method for aggregating the resultant basin characteristics for flow statistics calculations."

### LINKED ISSUE #1828: Combine basin characteristics for multiple watershed delineations
This is our temporary solution for computing basin characteristics for multipoint delineation (delineation by line). 

Multiple basins will be delineated and have basin characteristics computed. For this issue, mathematically combine all the basin characteristics to result in a single set of basin characteristics.

Pete M. can help with computations as needed.

amedenblik:
Harper did most of the work on #1169. I brought in my AM-1828 branch, connected the delineation and BC computation workflow, and started addressing bugs.

Differences between HW-1169 and this branch: https://github.com/USGS-WiM/StreamStats/compare/HW-1169...HW-1169-AM-1828

How things are working now:

1. You can draw a line to delineate multiple basins
2. The basins are merged. The merged basin is sent through to the NSS bylocation endpoint to get Regression Regions, Flow Statistics, and Basin Characteristics.
3. Each individual subbasin computes its basin characteristics separately. Then they are combined. This currently **only** works if all BCs of all subbasins were returned from the services.
4. The Flow Statistics are computed with the combined basin characteristics.


amedenblik:
Tasks addressed so far:

- [x] Don't show the WorkspaceID in reports for line delineation (already addressed this in the CSV)
- [x] Need to deal with the situation where some/all BCs aren't returned for some/all basins -- I have tried to fix that but can't seem to recreate the errors for further testing!
- [x] Need to address "only compute missing characteristics" section of the loadParameters function -- I have tried to address this but can't seem to recreate the errors for further testing
- [x] Some basin characteristics require other basin characteristics in order to compute them. Make sure these basin characteristics are computed. 
      - Any BC that uses "Area Weighted Mean" requires DRNAREA
      - CSL10_85fm requires LFPLENGTH
      - Then remove these from the sidebar and final report 
- [x] There are errors if the line doesn't intersect any streams
- [x] When you click "Continue" on the sidebar and it sends the polygon to the NSS bylocation endpoint, a toast appears about the "simplified basin"-- this should be turned off in this case, because we are always just using non-simplified basins for delineation by line
     - [x] This also occurs when you open the Report 
- [x] Sub-basins currently appear on the map by default. They should be turned off by default. The user will still have the option to turn them on via the legend.
- [x] We should probably refine some of the text in toasts, etc.
- [x] If you Delineate by Line and then click Clear Basin, only the Delineate by Line button is enabled-- the Delineation button is disabled. They should probably both be enabled? 
- [x] In the Report, subbassins appear as "globalwatershedX" when they should be called "Subbasin X"
- [x] Need to round values because ArcHydro isn't rounding them with this method

Remaining tasks are in #1852

harper-wavra (approved):

amedenblik:
Deployed to https://dev.streamstats.usgs.gov/sc-runoff/

Issue #1849, opened 2023-11-29
Disable DC catch basins below zoom level 16


amedenblik:
"I'm planning to approve the DC stormdrain QAQC for migration today but wanted to check on changing some view settings. Can we leave the storm drain inlets turned off until we get to a zoom level of 16 or higher? The inlets completely block all of the other map features and I don't think they're pertinent until you're zoomed way in and looking at very local delineations. I suspect we could do this with all of the stormdrain applications so we don't have custom setups for each one."

amedenblik:
Temporary fix: set minZoom to 14 for DC stormdrain layers

harper-wavra:
We figured out how to do this with just the map services. Waiting to hear if we should do the same with Mystic.

amedenblik:
Change was also made to the Mystic.

PR #1855, opened 2023-12-01
#1849 set DC stormdrain minZoom to 14

amedenblik:
This is a quick fix for #1849 before we have time to work on the actual requested feature.

Washington, D.C. Stormwater region StormDrainPipes layer is now set to minZoom 14. 

I went ahead and made this change on DevWeb.


harper-wavra:
We figured out how to do this with the map services.

PR #1884, opened 2023-12-18
Fixes #1883 - Bugs with delineate buttons

### LINKED ISSUE #1883: Delineate and Delineate By Line Button bugs
- [x] If you click "Delineate by Line" and immediately click "Delineate by Line", then click "Delineate", the line drawing interface is still active. Point delineation does still work though.
- [x] The button colors were not working as expected
- [x] Update the instruction text for the different options
- [x]  The spin icons in both the "Delineation" and "Delineate by Line" buttons spin while the delineation is happening. Might look nice if only the relevant one is spinning.

harper-wavra:
- When unselecting either of the delineation buttons, the cursor turns back into a pointer.
- When you select either of the options the other option disappears. If you selected the wrong button, you can unselect it before delineating & both options will reappear. I think this looks a bit cleaner and gets rid of the issue where both buttons were spinning when delineating.
- Updated the instruction text for the different options


amedenblik (approved):
Looks good! Good idea to hide the other delineate button after selection.

Issue #1852, opened 2023-11-30
Finish line delineation


amedenblik:
See #1169 and #1828 

Remaining tasks: (can create new issue(s))

**Essential**

- [x] I can't get basins to load when there are more than 2 intersection points. See [comment](https://github.com/USGS-WiM/StreamStats/issues/1852#issuecomment-1861633045).
- [x] If clicking second point too fast the line doesn't stop after two clicks #1903 
- [x] Need to deal with "additional features" (like longest flow path) #1901
    -  For longest flow path, use the longest of the individual basins.
- [x] Include an upper limit on the length of the line. ~~This could be by distance, in which case we could reuse code from the measure tool (client-side), or by the number of intersection points (service change). May need Pete's input on this.~~ https://github.com/USGS-WiM/StreamStats/pull/1902
    - [x] 10 intersections and 
    - [x] 2.5 miles 
- [x] Delete console.logs #1903 
- [ ] Improve Google analytics -- not sure if this is needed or what could be added?
     - Added Delineation Line to the [StreamStats Google Analytics Event Tracker](https://doimspp.sharepoint.com/:x:/r/sites/GS-UMidWIM/Shared%20Documents/Projects/Streamstats%20Ecosystem/StreamStats%20Google%20Analytics%20Event%20Tracker.xlsx?d=w759edfb901a648cb807e1f65561fde80&csf=1&web=1&e=h7ZP5I)
- Check that all aspects of the Report and the report downloadables (CSV, shapefile, etc.) still work #1902 
    - [x] Check download basin options [image](https://github.com/USGS-WiM/StreamStats/assets/59063181/284f55bd-b171-46ec-af54-4374597887b6) 
    - [x] Check report download. This is just spinning currently. 
- [x] Make sure Regulation feature works
- [x] Make sure Runoff Models feature still works with point delineation #1901 
- [x] Disable Runoff Models for line delineation #1901
- [x] Do testing to ensure all normal (point delineation) workflows still work
- [x] Disable NHD stream query for line delineation #1901
- [x] If you do a delineation, and then click Clear Basin, only one of the Delineate buttons appear (either Delineate or Delineate by Line, whichever was used). Then if you click that Delineate button, the other Delineate button does appear again and is disabled.  https://github.com/USGS-WiM/StreamStats/pull/1902
[...]

PR #1897, opened 2024-01-08
Fixes #1889 - Pull in Dev

### LINKED ISSUE #1889: Pull dev into SC-Runoff

harper-wavra:
Pulled in dev & rebuilt. There was a lot, we might want to pull in more frequently now that we are nearing the end. 
Some updates to make delineate by a line work if you select 'ignore exclude polygons'


amedenblik (approved):
Checked the following functionality:

- Point delineation in SC
- Point delineation outside SC
- Line delineation in SC
- Point delineation in SC, ignore exclude polys
- Line delineation in SC, ignore exclude polys

Issues I added to #1852:

- If you do a delineation, and then click Clear Basin, only one of the Delineate buttons appear (either Delineate or Delineate by Line, whichever was used). Then if you click that Delineate button, the other Delineate button does appear again and is disabled.
- If you draw a line in a hard exclude poly, it fails. Then you can try again and draw another line, but the old line still appears. The old line should disappear once a new line is drawn.

PR #1901, opened 2024-01-09
#1904, Part of #1852 - longest flow path, disable extra functions, fix storm runoff models

### LINKED ISSUE #1904: Line Delineation Report Download
The Download button in the report is always disabled after doing a line intersection. Once we get this button to be clickable again, we should check that the following download options all work as expected. 

- [x] GeoJSON
- [x] CSV
- [x] Shapefile
- [x] KML

amedenblik:
Completes parts of #1852 

- Handles longest flowpath functionality for line delineation
- Closes #1904
- Disable Runoff Models for line delineation
- Replaced Pourpoint.Latitude/Longitude with Pourpoint[0].Latitude/Longitude throughout the app
- Disable the Hydrologic Features functionality when doing line delineation


harper-wavra:
- [x] Sometimes I am not able to Continue after the Select Scenarios step. I click Continue & see null for all BCs, then see this [error](https://github.com/USGS-WiM/StreamStats/assets/59063181/5df684a8-a7db-4db0-9d4d-b47acf015a7f), and the sidebar does not proceed to the next step. If you are not seeing this, I can take a closer look on my end. 
- [x] Cannot download a csv from the report. The error has to do with trying to get the NHD streams, when there are none to get during a line intersection.

amedenblik:
Ready for re-review

- I changed the location of `this.eventManager.RaiseEvent(Services.onAdditionalFeaturesLoaded, this, '')` so that it only gets triggered once, when all additional features for all sub-basins have been loaded. This resolves the error message you were seeing.
- Fixed the CSV problems-- Hydrologic Features are only considered for point delineation
- Also removed an extra "Time" row in the CSV
- Let me know if you are still having issues with the "Continue" button. I wasn't able to recreate this after my changes.
- Pulled in latest from SC-Runoff

harper-wavra:
Tried with line and point delineation, it seemed like sometimes the longest flow path would download and sometimes it would not. Also seems to be returning the shorter of the two: image

amedenblik:
Unable to recreate errors on my end. We are going to try deploying to dev and see how it works there.

harper-wavra (approved):

Issue #1562, opened 2023-03-02
Streamline/Separate Gage Plots Loading 


maggiejaenicke:
- Currently, the modal tab "Gage Analysis Plots" (part of the gage page) freezes up for a second when clicked, and then the plots are all loaded simultaneously. We would like to have the tab open right when it is clicked, and then let the plots continue to load. Ideally, we would also separate the functions that go into creating each plot so they can load individually. The plots share the request for daily flow data.

- Could also just have one graph load quickly (maybe stage vs. discharge?), so the user gets some feedback that things are actually loading, while the others still load

- Additionally, in the Annual Peak Flow plot, gages with large data gaps in their daily flow data seem to be particularly slow (ex: 12340000). The daily flow array is being looped through so that each day between the start and end have a data row, and days without recordings get a null value to allow for a gap in the line. Maybe there is a better/quicker way to do this?

     - Nulls are added with this line: 

       `differences.forEach(date => this.formattedDailyFlow.push({x: date, y: null}))
`

**Tips:** 
- See issue https://github.com/USGS-WiM/StreamStats/issues/1424, issue https://github.com/USGS-WiM/StreamStats/issues/1499, and https://github.com/USGS-WiM/StreamStats/issues/1534 for details on the creation of these plots.
- See comments in pull request #1558 
- Make new branches based on the gageplots branch
- Editing will be done in the GagePageController.ts and the gagepage.html view

amedenblik:
We are seeing more some slowness now with the addition of the `updateFloodStats()` function. Some things to look into:

* Think about the efficiency of the 3 functions that get called: updateshadedstats, updatepeaksafterzoom, updatefloodstats 

* Does Highcharts have any reactive capabilities? Could we reference the extremes for the min/max x values of the horizontal lines?

* Does Highcharts have a nicer way to handle data gaps? How does weather.gov handle data gaps? 

* Maybe reconsider having so many plots loading at the same time.

PR #1913, opened 2024-01-23
Fixes #1707 Gageplots Feedback - Annual Peak Streamflow plot

### LINKED ISSUE #1707: Gageplots Feedback - Annual Peak Streamflow plot

maggiejaenicke:
Closes #1707

### DIFF
--- a/src/Controllers/GagePageController.ts
+++ b/src/Controllers/GagePageController.ts
@@ -901,6 +901,7 @@ module StreamStats.Controllers {
             } else {
                 url = 'https://nwis.waterservices.usgs.gov/nwis/dv/?format=json&sites=' + this.gage.code + '&parameterCd=00060&statCd=00003&startDT=1900-01-01&endDT=' + today;
             }
+            //console.log(url);
             this.dailyFlowURL = url;
             const request: WiM.Services.Helpers.RequestInfo = new WiM.Services.Helpers.RequestInfo(url, true, WiM.Services.Helpers.methodType.GET, 'json');
             this.Execute(request).then(
@@ -1447,8 +1448,7 @@ module StreamStats.Controllers {
             while (defaultYearPlusOne >= this.startAndEnd[0].getUTCFullYear()) {
                 this.allYears.push(defaultYearPlusOne -= 1)
             }
-            let endYear = endDate.getUTCFullYear();
-            let endOfFinalYear = new Date(9 + '/' + 30 + '/' + endYear)            
+            let endOfFinalYear = new Date(9 + '/' + 30 + '/' + this.defaultYear)            
             if (this.oneDayStats) {
                 this.formattedOneDayStats = [];
                 const oneDayStatsColors = {
@@ -1480,7 +1480,8 @@ module StreamStats.Controllers {
                             enabled: true,
                             zIndex: 3,
                             pointFormatter: function() { 
-                                if (this.x.getUTCFullYear() == endDate.getUTCFullYear()) {
+                                if (this.x) {
+                                    console.log(this.x, endOfFinalYear)
                                     return formattedName
                                 }
                             } 
@@ -1537,7 +1538,7 @@ module StreamStats.Controllers {
                             enabled: true,
                             zIndex: 3,
                             pointFormatter: function() { 
-                                if (this.x.getUTCFullYear() == endDate.getUTCFullYear()) {
+                                if (this.x) {
                                     return formattedName
                                 }
                             } 
@@ -1591,7 +1592,7 @@ module StreamStats.Controllers {
                             enabled: true,
                             zIndex: 3,
                             pointFormatter: function() { 
-                                if (this.x.getUTCFullYear() == endDate.getUTCFullYear()) {
+                                if (this.x) {
                                     return formattedName
                                 }
                             } 
@@ -1648,7 +1649,7 @@ module StreamStats.Controllers {
                             enabled: true,
                             zIndex: 3,
                             pointFormatter: function() { 
-                                if (this.x.getUTCFullYear() == endDate.getUTCFullYear()) {
+                                if (this.x) {
                                     return formattedName
                                 }
                             } 
@@ -1702,7 +1703,7 @@ module StreamStats.Controllers {
                             enabled: true,
                             zIndex: 3,
                             pointFormatter: function() { 
-                                if (this.x.getUTCFullYear() == endDate.getUTCFullYear()) {
+                                if (this.x) {
                                     return formattedName
                                 }
                             } 
@@ -1756,7 +1757,7 @@ module StreamStats.Controllers {
                             enabled: true,
                             zIndex: 3,
                             pointFormatter: function() { 
-                                if (this.x.getUTCFullYear() == endDate.getUTCFullYear()) {
+                                if (this.x) {
                                     return formattedName
                                 }
                             } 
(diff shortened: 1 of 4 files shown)


src/Controllers/GagePageController.ts
```diff
@@ -2986,6 +3212,84 @@ public chooseFloodStats() {
...
+            });
+        }
+        formattedSelectedPeaks.sort((a, b) => a.x - b.x);
+        formattedEstSelectedPeaks.sort((a, b) => a.x - b.x);
+        chart.series[0].update({data: formattedSelectedPeaks});
+        chart.series[1].update({data: formattedEstSelectedPeaks});
+        chart.yAxis[0].setExtremes();
+        chart.xAxis[0].setExtremes(min, max)
+        chart.series[0].update( {tooltip: {
+            headerFormat:'<b>Annual Peak Streamflow</b><br> Plotted on Water Year  HELLOOOOO: <b>' + this.selectedYear,
```
harper-wavra:
```suggestion
            headerFormat:'<b>Annual Peak Streamflow</b><br> Plotted on Water Year: <b>' + this.selectedYear,
```

src/Controllers/GagePageController.ts
```diff
@@ -889,13 +878,48 @@ module StreamStats.Controllers {
...
+			timeInMillisec -= 15 * 24 * 60 * 60 * 1000; 	// 14 days in milliseconds
+			date.setTime (timeInMillisec);
+            var twoWeeksAgo = new Date(date.getTime() - (date.getTimezoneOffset() * 60000 )) //ending date range two weeks ago to replace with IV
+                    .toISOString()
+                    .split("T")[0];
+            let url;
+            if (this.instFlow.length > 0) {
+                url = 'https://nwis.waterservices.usgs.gov/nwis/dv/?format=json&sites=' + this.gage.code + '&parameterCd=00060&statCd=00003&startDT=1900-01-01&endDT=' + twoWeeksAgo;
+            } else {
+                url = 'https://nwis.waterservices.usgs.gov/nwis/dv/?format=json&sites=' + this.gage.code + '&parameterCd=00060&statCd=00003&startDT=1900-01-01&endDT=2024-01-11';
```
harper-wavra: Should the `endDT` be hard coded here?

harper-wavra (changes requested):
Working well! This was a lot of changes, great job!

- Sometimes the log scale is not an option, [image](https://github.com/USGS-WiM/StreamStats/assets/59063181/f8c4eab8-4e5b-4780-91cf-9a980a3972db). I could not remember if we had talked about this, so maybe this is by design?
- After editing the time period shown on the plot, then clicking 'reset zoom' the page would sometimes freeze for me. This happened the most at gage 1234000. Not sure how easy this will be to fix, so could always make a separate issue.

src/Controllers/GagePageController.ts
maggiejaenicke: Lolll

maggiejaenicke:
> * Sometimes the log scale is not an option, [image](https://github.com/USGS-WiM/StreamStats/assets/59063181/f8c4eab8-4e5b-4780-91cf-9a980a3972db). I could not remember if we had talked about this, so maybe this is by design?

Similar to how the heat map determines if it should use log or linear, the log scale check box is only available if the data values do not contain any negative values (which causes the log option not to work on Highcharts). I have it set to not show the checkbox at all in this case (`ng-if="vm.containsNegatives() === false"`), but maybe it would be better/more self-explanatory if it were grayed out and still visible? 


> * After editing the time period shown on the plot, then clicking 'reset zoom' the page would sometimes freeze for me. This happened the most at gage 1234000. Not sure how easy this will be to fix, so could always make a separate issue.

12340000 seems to be super slow for a lot of things, and I think that is because of the huge data gap in the daily value data. The way that gap works is all the values are looped through so that each day has a data row and days without data get a null value. It does have to loop through a lot of data but I don't really know why it's soooo slow. I added some context about this to [this issue](https://github.com/USGS-WiM/StreamStats/issues/1562), so maybe we can work on address loading issues all at once.

harper-wavra (approved):
> Maybe it would be better/more self-explanatory if it were grayed out and still visible
I like this idea. 

Other than that, things are looking good to me!

amedenblik (changes requested):
Awesome job, Maggie! This is looking sharp.

- [x] On gage 03132000, I am not seeing labels for any of the the horizontal streamgage statistics lines:
image

- As discussed, it would be nice to improve loading times for gages that are especially slow, but we can address this in other issues.

maggiejaenicke:
Notes to self: 

> * [x]  On gage 03132000, I am not seeing labels for any of the the horizontal streamgage statistics lines:

I think I could just set this up in a function similar to `updateShadedStats()` or `updatePeaksAfterZoom()` to update the endpoints of the flood stats lines whenever the axis is resized. Would need this to apply to all of the flood stats, even the ones not currently shown, so that if they are switched in the dropdown, they have the right endpoints.

amedenblik (approved):

amedenblik:

PR #1940, opened 2024-02-01
Closes #1917 - MN Sediment - Add date range selector to modal

### LINKED ISSUE #1917: MN Sediment - Add date range selector to modal
Parent issue is #1544

## User Story

(continued from previous issue)

1. User is looking at the "Sediment Machine Learning Method" modal
2. Users selects a date range

## Tasks

- [x] Create a section of the modal that is titled "Select/Enter a Date Range (Required)"
- [x] At the end of a title, include a question mark icon that shows this text when the user hovers: "Date range is limited to 2006 and later due to model limitations."
- [x] Add a date range selector.
- [x] Prevent users from selecting dates before 2006-01-01
- [x] Include "Cancel" and "Ok" buttons at the bottom of the modal
- [x] The "Cancel" button should close the modal
- [x] The "Ok" button should be disabled unless all required elements are completed. So far, the only required element will be the date range selector.
- [x] When the user clicks the "Ok" button, the modal should close and a success toast should appear: "Sediment Machine Learning Method was successfully configured" / "Please continue"
- Hint: Use the "Flow Anywhere Model" modal and toast as an example

njestes:


src/Views/sedimentview.html
```diff
@@ -12,15 +12,14 @@ <h4 class="modal-title">Sediment Machine Learning Method</h4>
     <div class="modal-body">
         <div class="panel">
             <!-- https://github.com/fragaria/angular-daterangepicker -->
-            <!-- <small>Example to be changed for sediment model</small>
-            <h4 class="panel-title">Select/Enter a Date Range<small style="color:red">&nbsp;&nbsp;(Required)</small></h4>
-            <input id="dateSimple" date-range-picker class="form-control date-picker" type="text" ng-model="vm.dateRange.dates" min="vm.dateRange.minDate" max="vm.dateRange.maxDate" options="vm.dateRangeOptions" /> -->
+            <h4 class="panel-title">Select/Enter a Date Range<small style="color:red">&nbsp;&nbsp;(Required) &nbsp;&nbsp;</small><span><i title="Date range is limited to 2006 and later due to model limitations." class="fa fa-question-circle"></i></span></h4>
```
amedenblik, line 15: For consistency across the client, can we use this tooltip method?

Example: https://github.com/USGS-WiM/StreamStats/blob/dev/src/Views/batchprocessor.html#L9L12

You can view in the client by clicking the Batch Processor button at top right and hovering over the icon on the header:

image

amedenblik: For consistency across the client, can we use this tooltip method?

Example: https://github.com/USGS-WiM/StreamStats/blob/dev/src/Views/batchprocessor.html#L9L12

You can view in the client by clicking the Batch Processor button at top right and hovering over the icon on the header:

image

src/Controllers/SedimentController.ts
```diff
@@ -119,6 +128,29 @@ module StreamStats.Controllers {
             //     }
             // }
         }
+        private verifyExtensionCanContinue() {
```
amedenblik, line 131: I was able to enter this invalid date range by typing it in and tab-ing over to the Ok button:

image

The Ok button just blinked and didn't give a success message. I got this console error:

image

Maybe we should do some additional error checking to make sure startDate and endDate are defined?

amedenblik (changes requested):

src/Controllers/SedimentController.ts
njestes: Interestingly, when out of min/max range (even if just one or the other) set on the input, the `this.dateRange.dates` (where Start and End date are found) property is left entirely undefined. So, a simple fix could be that I just check that that is defined and if not, still return the same "Date range is not valid." error. Or would we prefer for it be more specific about whether it's Start Date and/or End Date that is not defined properly?

amedenblik: > a simple fix could be that I just check that that is defined and if not, still return the same "Date range is not valid." error

That sounds good to me

njestes: It may be as simple as checking `if (this.dateRange.dates) {` rather than just `if (this.dateRange) {` which it is currently doing as the very first check

amedenblik: Maybe `if (this.dateRange.dates.startDate && this.dateRange.dates.endDate) {`

njestes: ok, attempt made to address the last review

amedenblik (approved):
Changes look good 👍

PR #1956, opened 2024-02-14
Fixes #1953 - enable FL

### LINKED ISSUE #1953: Enable Florida
- [x] enable Florida in the appconfig (currently just delineation) 
- [x] Pete S would like the imagery base map displayed by default. We should just have to add 'localRes' to the Applications in the appconfig for Flordia.

harper-wavra:
Enabled FL in the Dev app config, added localres to FL applications

Once working on dev will change out of draft status.


amedenblik:
I think we could go ahead and merge this, since then the appConfig_dev.js will match what is actually on DevWeb.

harper-wavra:
Should we remove the changelog item then? Since that would end up going to production? Or I guess I specified 'dev' maybe that's fine.

amedenblik (approved):

Issue #1880, opened 2023-12-15
Add note about typescript version


harper-wavra:
It seems we are no longer able to run `npm install -g typescript` when getting the repo up and running locally. I was getting typescript errors when trying a fresh clone, I think due to an old version of nodejs that didn't support "??" operator yet. See  https://stackoverflow.com/questions/76421238/tsc-command-showing-syntaxerror-unexpected-token.

When I ran npm install typescript@4.9.4 it worked. 

We should either add a note to the readme or else add typescript to the package.json

harper-wavra:
Might just need to add "typescript": "^4.9.4" to the package.json

PR #1961, opened 2024-02-21
Gageplots - fix typescript installation

amedenblik:
Apply fix for #1880 (#1882) to the gageplots branch


harper-wavra:
I wonder if we should just pull in dev instead? I have noticed things get kinda weird when editing the same thing in two branches.

amedenblik:
Good call. Will do!

PR #1974, opened 2024-02-28
Fixes #1925 Stage vs. Discharge Plot - Fix plot scale so all values are shown in linear scale

### LINKED ISSUE #1925: Stage vs. Discharge Plot - Fix plot scale so all values are shown in linear scale
Sometimes the scale on the stage vs. discharge plots cuts off some values. 
The default scale for 06125520 cuts off all but two values. Switching to the log scale shows the other values.
All values are also cut off on 05513450 & 03419000, but are shown when switched to log scale.

- [x] Fix plot scale so all values are shown when plot is in linear scale

Replaces item in #1708

maggiejaenicke:
Plots that only have peak data were having an issue with the linear scale because the yAxis `tickPositioner` for the Stage vs. Discharge plot was only using the USGS Measured data to gather min and max. I added a check for peak data values, which for some of these gages was nonexistent, so now it uses whichever data is available or the overall min/max. 

Gages to test that previously showed nothing in linear scale: 

- 05513450
- 03419000
- 06125520

Closes #1925

### DIFF
--- a/src/Controllers/GagePageController.js
+++ b/src/Controllers/GagePageController.js
@@ -3051,6 +3051,26 @@ var StreamStats;
                 var measuredDataMin = Math.min.apply(Math, this.measuredObj
                     .filter(function (obj) { return typeof obj.y === 'number' && !isNaN(obj.y); })
                     .map(function (obj) { return obj.y; }));
+                var peakDataMax = Math.max.apply(Math, this.formattedDischargePeakDates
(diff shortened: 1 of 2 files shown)


amedenblik (approved):
Works well!

amedenblik:
Deployed to https://dev.streamstats.usgs.gov/gageplots

PR #1977, opened 2024-03-12
Fixes #1971: Gageplots - fix Print appearance, Fixes #1976: Add note about toggling data series

### LINKED ISSUE #1971: Gageplots - fix Print appearance
Issue from #1709 

 - [x] Fix small issues in appearance when you click the "Print" button at bottom of modal
      - All flood stats lines are showing up -- we want it to just be the selected lines
      - separate plots into different pages with forced page breaks

### LINKED ISSUE #1976: Add note about toggling data series
Underneath the "Click and drag..." note on each plot, add a line that says "Click data series name in legend to toggle on/off" or similar language-- feel free to workshop. Pete did not realize that you could turn on/off layers by clicking their name in the legend, so I'm sure other users would not find this intuitive either.

maggiejaenicke:
Closes #1971
- Adjusts printer CSS
- Adds forced page breaks between plots in the HTML
- Fixes issue of ALL flood stats showing up when page printed, now it should just be the selected ones

 Closes #1976 
- Adds note about toggling data series to each plot

### DIFF
--- a/src/Controllers/GagePageController.js
+++ b/src/Controllers/GagePageController.js
@@ -2886,6 +2886,7 @@ var StreamStats;
                 var chart = $('#chart1').highcharts();
                 chart.showLoading('Loading...');
                 setTimeout(function () {
+                    _this_1.updateFloodStats();
                     var extremes = chart.xAxis[0].getExtremes();
                     var min = new Date(extremes.min);
                     var max = new Date(extremes.max);
@@ -2917,7 +2918,6 @@ var StreamStats;
                     }
                     chart.hideLoading();
                 }, 100);
-                this.updateFloodStats();
             };
             GagePageController.prototype.choosePeakYear = function () {
                 var _this_1 = this;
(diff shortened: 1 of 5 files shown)


amedenblik (changes requested):
As discussed, there's something off about the month (21) in the request URL (https://nwis.waterservices.usgs.gov/nwis/dv/?format=json&sites=03419000&parameterCd=00060&statCd=00003&startDT=1900-01-01&endDT=2024-21-13), so we're not able to load some plots.

I reviewed the changes to address #1971 and #1976 on plots that were able to load (ex. http://127.0.0.1:8080/src/?gage=02173500&tab=plots), and they look good!

maggiejaenicke:
Should be fixed! It was just adding the month + 1 as a string for some reason instead of an int. Good catch--most plots have instantaneous values so the url uses a date from two weeks ago instead of the current date, which is why most plots are fine. It was likely just erroring for any plots that don't have IV.

amedenblik (approved):
Now working at all gages!

amedenblik:

PR #1979, opened 2024-03-13
Fixes #1978

### LINKED ISSUE #1978: Rural Hydrograph computational error
SCDOT reports that USGS rural hydrograph computations for the 2% (50yr) and 1%(100 yr) hydrographs are much larger than expected.
[34.44148, -81.03518](https://github.com/USGS-WiM/StreamStats/assets/59063181/34cb97d0-7ed2-49d5-9353-9577cc749a69)

Issue seems to be with lines https://github.com/USGS-WiM/StreamStats/blob/SC-Runoff/src/Controllers/SCStormRunoffController.ts#L844-#L848, where if the AEP is 1 or 2, its also grabbing the results from 10 and 20 AEP results.

harper-wavra:
- Changed to use code to grab the correct results
- Was happening with Urban results so made the change there as well

To test I used the points in the Word document Kitty sent, 34.44148, -81.03518 and 33.96926, -79.18598 then compared the results to the chart that was provided in the document. 

We only want to include PKXXAEP correct?

### DIFF
--- a/src/Controllers/SCStormRunoffController.js
+++ b/src/Controllers/SCStormRunoffController.js
@@ -626,10 +626,7 @@ var StreamStats;
                             response.data[0].regressionRegions.forEach(function (regressionregion) {
                                 if (regressionregion.code == 'GC1583' || regressionregion.code == 'GC1584' || regressionregion.code == 'GC1585' || regressionregion.code == 'GC1586') {
                                     regressionregion.results.forEach(function (result) {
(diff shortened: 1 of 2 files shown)


amedenblik:
Original issue: https://code.usgs.gov/StreamStats/web-services-and-apis/SC-RunoffModelingServices/-/issues/110

amedenblik (approved):
Looks good to me! Yes, I believe we only want to include PKXXAEP.

